PR #1, opened 2019-10-15
Add black box design proposal

alexmt:
PR contains GitOps engine design proposal which references two documents:
 - design-black-box.md - describes the POC approach
 - design-white-box.md - an empty document for now. In next PR will move the content of https://docs.google.com/document/d/10OvgB9nMmleYPxPlLuXq6RNFodJSKZN7GUSNTyfwEhw/edit#

### DIFF
--- a/specs/design-black-box.md
+++ b/specs/design-black-box.md
@@ -33,7 +33,12 @@ be wrong. So if we are missing some important Flux feature the approach might be
 
 ### Risks and Mitigations
 
-To mitigate the risk of let's start from POC to catch possible blockers earlier. 
+The proposed solution is based on the assumption that despite implementation differences the core functionality of Argo CD and Flux behaves in the same way. Both projects
+ultimately extract the set of manifests from Git and use "kubectl apply" to change the cluster state. The minor differences are expected but we can resolve them by introducing new
+knobs.
+
+However, there is a risk that there will be two many differences and it might be not feasible to support all of them. To mitigate the risk of let's start from POC implementation
+to catch possible blockers earlier. 
 
 ### GitOps Engine POC
 
--- a/specs/design-black-box.md
+++ b/specs/design-black-box.md
@@ -1,0 +1,58 @@
+# GitOps Engine Design - Black Box
+
+## Summary
+
+During the elaboration of [White box](./design-white-box.md) option, it was discovered that some components are similar at a high-level but still have a lot of differences in 
+design and implementation. This is not surprising because code was developed by different teams and with a focus on different use-cases. Given that it would require a lot of
+effort to resolve such differences it is proposed contributing missing features of one project into an engine of another and use that engine in both projects.
+
+## Proposal
+
+It is proposed to use Argo CD application controller as the base for the GitOps engine and contribute a set of Flux features into it. There are two main reasons to try using Argo CD as
+a base:
+- Argo CD uses Application abstraction to represent the desired state and target Kubernetes cluster. This abstraction works for both Argo CD and Flux cases.
+- The Argo CD controller leverages Kubernetes watch APIs instead of pulling. This enables Argo CD features such as Health assessment, UI and could provide better performance to
+Flux as well.
+
+Following Flux features are missing in Argo CD:
+
+- Manifest generation using .flux.yaml files.
+- GPG commit signatures verification - an ability to verify the commit signature before pushing changes to the Kubernetes.
+- Namespace mode - an ability to control only given namespace in the target cluster. Currently, Argo CD requires read access in all namespaces.
+
+These features must be contributed to Argo CD before moving Argo CD controller into Argo Flux GitOps engine repo.
+
+Flux additionally provides the ability to monitor Docker registry and automatically push changes to the Git repository when a new image is released. Both teams feel the this should not
+be a part of GitOps engine. So it is proposed to keep the feature only in Flux for now and then work together to move it into a separate component that would work for both Flux
+and Argo CD.
+
+### Implementation Details/Notes/Constraints
+
+The proposed approach is based on the assumption that Argo CD engine is flexible enough to cover all Flux use-cases and can be easily integrated into Argo CD. This assumption might
+be wrong. So if we are missing some important Flux feature the approach might be no feasible.
+
+### Risks and Mitigations
+
+To mitigate the risk of let's start from POC to catch possible blockers earlier. 
+
+### GitOps Engine POC
+
+POC deliverables are:
+
+- All POC changes are in separate branches.
+- Argo CD controller will be moved to https://github.com/argoproj/gitops-engine.
+- Flux will import GitOps engine component from the https://github.com/argoproj/gitops-engine repository and use it to perform cluster state syncing.
+- The flux installation and fluxctl behavior will remain the same other than using GitOps engine internally. That means there will be no creation of Application CRD or Argo CD
+specific ConfigMaps/Secrets.
+- For the sake of saving time POC does not include implementing features mentioned before. So no commit verification, only plain .yaml files support, and full cluster mode.
+
+## Design Details
+
+### Public API
+
+To be documented during POC implementation.
+
+
+## Alternatives
+
+[White box](./design-white-box.md)
--- a/specs/design-black-box.md
+++ b/specs/design-black-box.md
@@ -14,7 +14,7 @@ a base:
 - The Argo CD controller leverages Kubernetes watch APIs instead of pulling. This enables Argo CD features such as Health assessment, UI and could provide better performance to
 Flux as well.
 
-Following Flux features are missing in Argo CD:
+The following Flux features are missing in Argo CD:
 
 - Manifest generation using .flux.yaml files.
 - GPG commit signatures verification - an ability to verify the commit signature before pushing changes to the Kubernetes.
--- a/specs/design-black-box.md
+++ b/specs/design-black-box.md
@@ -37,7 +37,7 @@ The proposed solution is based on the assumption that despite implementation dif
 ultimately extract the set of manifests from Git and use "kubectl apply" to change the cluster state. The minor differences are expected but we can resolve them by introducing new
 knobs.
 
-However, there is a risk that there will be two many differences and it might be not feasible to support all of them. To mitigate the risk of let's start from POC implementation
+However, there is a risk that there will be too many differences and it might be not feasible to support all of them. To mitigate the risk of let's start from POC implementation
 to catch possible blockers earlier. 
 
 ### GitOps Engine POC
--- a/specs/design-black-box.md
+++ b/specs/design-black-box.md
@@ -20,7 +20,7 @@ The following Flux features are missing in Argo CD:
 - GPG commit signatures verification - an ability to verify the commit signature before pushing changes to the Kubernetes.
 - Namespace mode - an ability to control only given namespace in the target cluster. Currently, Argo CD requires read access in all namespaces.
 
-These features must be contributed to Argo CD before moving Argo CD controller into Argo Flux GitOps engine repo.
+These features must be contributed to Argo-Flux GitOps engine implementation before Flux starts using it.
 
 Flux additionally provides the ability to monitor Docker registry and automatically push changes to the Git repository when a new image is released. Both teams feel the this should not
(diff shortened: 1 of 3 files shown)


specs/design-black-box.md
```diff
@@ -0,0 +1,58 @@
...
+These features must be contributed to Argo CD before moving Argo CD controller into Argo Flux GitOps engine repo.
+
+Flux additionally provides the ability to monitor Docker registry and automatically push changes to the Git repository when a new image is released. Both teams feel the this should not
+be a part of GitOps engine. So it is proposed to keep the feature only in Flux for now and then work together to move it into a separate component that would work for both Flux
+and Argo CD.
+
+### Implementation Details/Notes/Constraints
+
+The proposed approach is based on the assumption that Argo CD engine is flexible enough to cover all Flux use-cases and can be easily integrated into Argo CD. This assumption might
+be wrong. So if we are missing some important Flux feature the approach might be no feasible.
```
2opremio: This is a great start, but I think we should expand it into an _Acceptance Criteria_ section indicating more explicitly what would constitute a successful PoC (or a failure).

2opremio: I can take that up if you want

specs/design-black-box.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Summary
+
+During the elaboration of [White box](./design-white-box.md) option, it was discovered that some components are similar at a high-level but still have a lot of differences in 
+design and implementation. This is not surprising because code was developed by different teams and with a focus on different use-cases. Given that it would require a lot of
+effort to resolve such differences it is proposed contributing missing features of one project into an engine of another and use that engine in both projects.
+
+## Proposal
+
+It is proposed to use Argo CD application controller as the base for the GitOps engine and contribute a set of Flux features into it. There are two main reasons to try using Argo CD as
```
2opremio: During or prior conversations I didn't understand that the PoC was going to use the controller _as is_, but that it would have some internal rewiring to cater Flux. There is nothing wrong with it per se, but it increases my concern about Flux keeping its current behavior.

Two good examples of such behavior are Flux' Syncing and Garbage collection (but there are more). The syncing process and GC of Flux are fine tuned and contain a lot of nuances. 

For instance, GC carefully uses labels used in a very specific way, which may already be used as an assumption by Flux users and impact when a resource can be garbage collected (see https://docs.fluxcd.io/en/stable/references/garbagecollection.html ). 

In terms of syncing, we are very careful about ordering, continuously reconciling the status of Git with the cluster and causing no harm (giving up in each iteration after seeing any errors ).

And what worries me the most is that, due to the complexity, we cannot have all those details in our heads when evaluating the PoC (causing problems down the line).

Please note I am not writing this to discourage writing the PoC, but to clarify what my concerns.

specs/design-black-box.md
```diff
@@ -0,0 +1,58 @@
...
+a base:
+- Argo CD uses Application abstraction to represent the desired state and target Kubernetes cluster. This abstraction works for both Argo CD and Flux cases.
+- The Argo CD controller leverages Kubernetes watch APIs instead of pulling. This enables Argo CD features such as Health assessment, UI and could provide better performance to
+Flux as well.
+
+Following Flux features are missing in Argo CD:
```
2opremio:
```suggestion
The following Flux features are missing in Argo CD:
```

specs/design-black-box.md
```diff
@@ -0,0 +1,58 @@
...
+
+During the elaboration of [White box](./design-white-box.md) option, it was discovered that some components are similar at a high-level but still have a lot of differences in 
+design and implementation. This is not surprising because code was developed by different teams and with a focus on different use-cases. Given that it would require a lot of
+effort to resolve such differences it is proposed contributing missing features of one project into an engine of another and use that engine in both projects.
+
+## Proposal
+
+It is proposed to use Argo CD application controller as the base for the GitOps engine and contribute a set of Flux features into it. There are two main reasons to try using Argo CD as
+a base:
+- Argo CD uses Application abstraction to represent the desired state and target Kubernetes cluster. This abstraction works for both Argo CD and Flux cases.
```
2opremio:
```suggestion
- Argo CD uses the _Application_ abstraction to represent the desired state and target Kubernetes cluster. This abstraction works for both Argo CD and Flux cases.
```

specs/design-black-box.md
```diff
@@ -0,0 +1,58 @@
...
+To mitigate the risk of let's start from POC to catch possible blockers earlier. 
+
+### GitOps Engine POC
+
+POC deliverables are:
+
+- All POC changes are in separate branches.
+- Argo CD controller will be moved to https://github.com/argoproj/gitops-engine.
+- Flux will import GitOps engine component from the https://github.com/argoproj/gitops-engine repository and use it to perform cluster state syncing.
+- The flux installation and fluxctl behavior will remain the same other than using GitOps engine internally. That means there will be no creation of Application CRD or Argo CD
```
2opremio: I think it's worth clarifying what _behavior_ means , the interface? the result (both in the cluster and the git repo when invoking them? To what extend?

specs/design-black-box.md
```diff
@@ -0,0 +1,58 @@
...
+and Argo CD.
+
+### Implementation Details/Notes/Constraints
+
+The proposed approach is based on the assumption that Argo CD engine is flexible enough to cover all Flux use-cases and can be easily integrated into Argo CD. This assumption might
+be wrong. So if we are missing some important Flux feature the approach might be no feasible.
+
+### Risks and Mitigations
+
+To mitigate the risk of let's start from POC to catch possible blockers earlier.
```
2opremio: Missing risk :)

2opremio:
Great effort and well phrased! I only have a few comments, mostly clarifying my position and asking for acceptance criteria for the PoC (which I am willing to write if you prefer) to avoid confusions later on

specs/design-black-box.md
```diff
@@ -0,0 +1,58 @@
...
+design and implementation. This is not surprising because code was developed by different teams and with a focus on different use-cases. Given that it would require a lot of
+effort to resolve such differences it is proposed contributing missing features of one project into an engine of another and use that engine in both projects.
+
+## Proposal
+
+It is proposed to use Argo CD application controller as the base for the GitOps engine and contribute a set of Flux features into it. There are two main reasons to try using Argo CD as
+a base:
+- Argo CD uses Application abstraction to represent the desired state and target Kubernetes cluster. This abstraction works for both Argo CD and Flux cases.
+- The Argo CD controller leverages Kubernetes watch APIs instead of pulling. This enables Argo CD features such as Health assessment, UI and could provide better performance to
+Flux as well.
```
2opremio: 👏

specs/design-black-box.md
```diff
@@ -0,0 +1,58 @@
...
+During the elaboration of [White box](./design-white-box.md) option, it was discovered that some components are similar at a high-level but still have a lot of differences in 
+design and implementation. This is not surprising because code was developed by different teams and with a focus on different use-cases. Given that it would require a lot of
+effort to resolve such differences it is proposed contributing missing features of one project into an engine of another and use that engine in both projects.
+
+## Proposal
+
+It is proposed to use Argo CD application controller as the base for the GitOps engine and contribute a set of Flux features into it. There are two main reasons to try using Argo CD as
+a base:
+- Argo CD uses Application abstraction to represent the desired state and target Kubernetes cluster. This abstraction works for both Argo CD and Flux cases.
+- The Argo CD controller leverages Kubernetes watch APIs instead of pulling. This enables Argo CD features such as Health assessment, UI and could provide better performance to
```
2opremio:
```suggestion
- The Argo CD controller leverages Kubernetes watch APIs instead of polling. This enables Argo CD features such as Health assessment, UI and could provide better performance to
```

specs/design-black-box.md
```diff
@@ -0,0 +1,58 @@
...
+### GitOps Engine POC
+
+POC deliverables are:
+
+- All POC changes are in separate branches.
+- Argo CD controller will be moved to https://github.com/argoproj/gitops-engine.
+- Flux will import GitOps engine component from the https://github.com/argoproj/gitops-engine repository and use it to perform cluster state syncing.
+- The flux installation and fluxctl behavior will remain the same other than using GitOps engine internally. That means there will be no creation of Application CRD or Argo CD
+specific ConfigMaps/Secrets.
+- For the sake of saving time POC does not include implementing features mentioned before. So no commit verification, only plain .yaml files support, and full cluster mode.
```
hiddeco: Is there a way to guarantee that this doesn't lead to a 80% working PoC, that gives you an amazing head start and/or confirmation that it would work in theory, but ends up with the last 20% taking a tremendous amount of time due to unforeseen complexities and/or looked over features?

specs/design-black-box.md
hiddeco: > During our prior conversations I didn't understand that the PoC was going to use the controller as is, but that it would have some internal rewiring to cater Flux. There is nothing wrong with it per se, but it increases my concern about Flux not being able to keep its current behavior.
> [...]
> And what worries me the most is that, due to the complexity, we cannot have all those details in our heads when evaluating the PoC (causing problems down the line).
> [...]
> Please note I am not writing this to discourage writing the PoC, but to clarify what my concerns are.

+1

specs/design-black-box.md
alexmt: sure, I think it would be better if you write it

specs/design-black-box.md
alexmt: Oops, looks like it was cut off during rewriting of other section. Fixed.

specs/design-black-box.md
alexmt: Both the interface and the result (taking into account known missing features). Can we cover these details in the Acceptance Criteria ?

specs/design-black-box.md
2opremio: Will do

specs/design-black-box.md
```diff
@@ -0,0 +1,63 @@
...
+- The Argo CD controller leverages Kubernetes watch APIs instead of polling. This enables Argo CD features such as Health assessment, UI and could provide better performance to
+Flux as well.
+
+The following Flux features are missing in Argo CD:
+
+- Manifest generation using .flux.yaml files.
+- GPG commit signatures verification - an ability to verify the commit signature before pushing changes to the Kubernetes.
+- Namespace mode - an ability to control only given namespace in the target cluster. Currently, Argo CD requires read access in all namespaces.
+
+These features must be contributed to Argo CD before moving Argo CD controller into Argo Flux GitOps engine repo.
```
jaypipes: Any reason why this is the case? Would it not be possible to move ArgoCD app controller into gitops-engine repo and add these features to the gitops-engine repo?

specs/design-black-box.md
```diff
@@ -0,0 +1,63 @@
...
+
+Flux additionally provides the ability to monitor Docker registry and automatically push changes to the Git repository when a new image is released. Both teams feel the this should not
+be a part of GitOps engine. So it is proposed to keep the feature only in Flux for now and then work together to move it into a separate component that would work for both Flux
+and Argo CD.
```
jaypipes: :+1:

specs/design-black-box.md
```diff
@@ -0,0 +1,63 @@
...
+The proposed approach is based on the assumption that Argo CD engine is flexible enough to cover all Flux use-cases and can be easily integrated into Argo CD. This assumption might
+be wrong. So if we are missing some important Flux feature the approach might be no feasible.
+
+### Risks and Mitigations
+
+The proposed solution is based on the assumption that despite implementation differences the core functionality of Argo CD and Flux behaves in the same way. Both projects
+ultimately extract the set of manifests from Git and use "kubectl apply" to change the cluster state. The minor differences are expected but we can resolve them by introducing new
+knobs.
+
+However, there is a risk that there will be two many differences and it might be not feasible to support all of them. To mitigate the risk of let's start from POC implementation
```
hiddeco:
```suggestion
However, there is a risk that there will be too many differences and it might be not feasible to support all of them. To mitigate the risk of let's start from POC implementation
```

specs/design-black-box.md
alexmt: I agree. The order does not really matter. Updated the sense.

PR #2, opened 2019-10-17
Add black box hypothesis and acceptance criteria

2opremio:


specs/design-black-box.md
```diff
@@ -26,25 +26,43 @@ Flux additionally provides the ability to monitor Docker registry and automatica
...
+   - Flux's API must remain unchanged. In particular, the web API (used by fluxctl) and the websocket API (e.g. used to 
+     communicate with Weave Cloud) must work without changes.
+3. Flux's writing behaviour on Git and Kubernetes must be identical. In particular:
+   - Flux+GitEngine should make changes in Git if and only if Flux without GitEngine would had done it,
+     in the same way (same content) and in the same situations
+   - Flux+GitEngine should add and update Kubernetes resources if and only if Flux without GitEngine would had done, 
+     in the same way (same content) and in the same situations
+
+Additionally, there must be a clear way forward (in the shape of well-defined steps) 
+for the features not covered by the PoC to work (complying with the points avoid) int he final GitOps
```
alexmt:
```suggestion
for the features not covered by the PoC to work (complying with the points avoid) in the final GitOps  
```

jaypipes: "complying with the points avoid" -> I think you meant "the points above"? Also, "int he final GitOps" should be "in the final GitOps"

specs/design-black-box.md
```diff
@@ -26,25 +26,43 @@ Flux additionally provides the ability to monitor Docker registry and automatica
...
+Also, the proposed approach is based on the assumption that Argo CD engine is flexible enough to cover all Flux use-cases, reproduce Flux's behavior with minor differences and can be easily integrated into Argo CD.
+
+However, there is a risk that there will be too many differences and it might be not feasible to support all of them. To get early feedback, we will start with a Proof-of-Concept 
+(PoC from now on) implementation which will serve as an experiment to asses the feasibility of the approach.
+
+### Acceptance criteria
+
+To consider the PoC successful (and with the exception of features excluded from the PoC to save time), 
+all the following must hold true:
+1. All the Flux unit and end-to-end tests must pass. The existing tests are limited, so we may decide to include additional ones.
```
hiddeco: Most of the unit tests we have for core functionalities are contained within the packages that provide the logic, as the engine replaces those packages, the tests will also no longer be there.

The (very basic) end-to-end test on the other hand only test the end result (I expect something to appear in the cluster, did this happen?), without looking into the details of how this is done.

Given the above I do not see how can assure acceptance criterion 3 without any additional tests, and I thus think tests covering this are a requirement from our side.

alexmt: This a good point. Assuming Argo CD and Flux are going to merge one way other another we would benefit from Flux e2e tests either way. Should we start working on e2e tests in parallel?

hiddeco: I think we can work on a framework to make e2e testing easy in parallel. But due to both applications having their own ways of doing things, the tests should be focused on how things are expected to be handled by an application _as is_. This will assure things still work as expected once the engine has replaced the business logic each application currently holds.

alexmt: I agree ( about verifying of current behavior ). Argo CD has e2e framework which might be useful. @hiddeco , @2opremio, please let @alexec know if you want to work on flux e2e tests and need help with it.

2opremio: > I think we can work on a framework to make e2e testing easy in parallel

I totally agree. I will work on that.

> Argo CD has e2e framework which might be useful. @hiddeco , @2opremio, please let @alexec know if you want to work on flux e2e tests and need help with it.

Thanks, I will take a look

specs/design-black-box.md
```diff
@@ -26,25 +26,44 @@ Flux additionally provides the ability to monitor Docker registry and automatica
...
+3. Flux's writing behaviour on Git and Kubernetes must be identical. In particular:
+   - Flux+GitEngine should make changes in Git if and only if Flux without GitEngine would had done it,
+     in the same way (same content) and in the same situations
+   - Flux+GitEngine should add and update Kubernetes resources if and only if Flux without GitEngine would had done, 
+     in the same way (same content) and in the same situations
+
+Unfortunately, there isn't a straightforward way to decidedly check for (3).
+
+Additionally, there must be a clear way forward (in the shape of well-defined steps) 
+for the features not covered by the PoC to work (complying with the points above) int he final GitOps
```
hiddeco:
```g
for the features not covered by the PoC to work (complying with the points above) in the final GitOps
```

2opremio: darn, we merged before I addressed this. I will fix in master

PR #3, opened 2019-10-18
Document core functionality differences

alexmt:
PR documents discovered differences in core Argo CD and Flux functionality with the proposed resolution category. 

Still WIP but early feedback is very appreciated.

### DIFF
--- a/specs/design-black-box.md
+++ b/specs/design-black-box.md
@@ -26,6 +26,35 @@ Flux additionally provides the ability to monitor Docker registry and automatica
 be a part of GitOps engine. So it is proposed to keep the feature only in Flux for now and then work together to move it into a separate component that would work for both Flux
 and Argo CD.
 
+### Differences in Core Functionality
+
+Even core functionality of Argo CD has some differences. These differences might look minor but affect production systems and cannot be ignored. It is proposed to document all such
+differences, document and propose a resolution. Following three categories are proposed:
+
+- `Feature`. The difference is justified by a real business requirement and affects the user. To resolve it we should contribute a feature into GitOps engine.
+- `Mode`. The is no real business requirement and difference exists because slightly different design decision was made during Argo CD/Flux implementation. To resolve it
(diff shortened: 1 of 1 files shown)


specs/design-black-box.md
```diff
@@ -26,6 +26,35 @@ Flux additionally provides the ability to monitor Docker registry and automatica
...
+we would have to introduce a setting that switches Engine between Argo CD/Flux mode. Going forward we should try to pick one behavior, deprecate the previous one and eventually migrate
+all users to one behavior.
+- `Random`. The difference is caused by an internal implementation detail and does not affect the user in any way. We can just ignore the difference.
+
+**Differences List (WIP)**:
+
+- `Feature` Policy annotation `fluxcd.io/ignore`. If either resource manifest in Git or live resource in the target cluster has an `fluxcd.io/ignore` then Flux doesn't touch it.
+Argo CD does not have a similar annotation.
+
+- `Feature` Garbage collection/resource pruning. Garbage collection logic in Argo CD and Flux is slightly different. Argo CD inject label `app.kubernetes.io/instance: <appName>`
```
alexmt: It seems like this can be covered by the existing GitOps engine functionality. The engine allows to dynamically choose value of applied label (so it can be changed  from `app.kubernetes.io/instance` and value and be set dynamically to `sha256.<checksum>`)

@2opremio , @hiddeco I will need to clarify how exactly `fluxcd.io/sync-checksum`  annotation is used.

hiddeco: The `sync-checksum` contains the checksum of the applied manifest, this is to be able to tell if a resource has been updated so that they aren't removed in case there hasn't been an update.

Most of the logic around this checksum can be found here: https://github.com/fluxcd/flux/blob/f1698542379402e67596c7c579a557ad0dd542bf/pkg/cluster/kubernetes/sync.go#L129

alexmt:
Looks like we can support both Argo CD and Flux behavior without contributing user facing features. So closing this PR

PR #9, opened 2019-10-23
describe license and code-of-conduct

dholbach:
Add a note about the license and code-of-conduct (using the one from CNCF) we use.


jaypipes:
merging this since I doubt this would be a source of controversy

PR #12, opened 2019-11-01
Add engine API to the black-box design

alexmt:
PR adds the target engine API.

The design is only inspired by PoC but represents that final state so pretty different from the implemented changes.

### DIFF
--- a/specs/design-black-box.md
+++ b/specs/design-black-box.md
@@ -121,7 +121,7 @@ type Engine interface {
 	// OnSyncCompleted registers callback that is executed after each sync operation.
 	OnSyncCompleted(callback func(appName string, state appv1.OperationState) error) Unsubscribe
 
-	// OnClusterCacheInitialized registers that that is executed when cluster cache initialization is completed.
+	// OnClusterCacheInitialized registers a callback that is executed when cluster cache initialization is completed.
 	OnClusterCacheInitialized(callback func(server string)) Unsubscribe
 
 	// OnResourceUpdated registers that that is executed when cluster resource got updated.
--- a/specs/design-black-box.md
+++ b/specs/design-black-box.md
@@ -77,22 +77,22 @@ The proposed design is based on the PoC that contains draft implementation and p
 
 The GitOps engine API consists of three main packages:
 * The `utils` package. It consist of loosely coupled packages that implements K8S resource diffing, health assessment, etc
-* The `app` package that Application data structure definition and CRUD client interface.
+* The `syncunit` package that contains SyncUnit data structure definition and CRUD client interface.
 * The `engine` package that leverages `utils` and uses provided set of Applications as a configuration source.
 
 ```
 gitops-engine
 |-- pkg
 |   |-- utils
-|   |   |-- diff   # provides Kubernetes resource diffing functionality
-|   |   |-- kube   # provides utility methods to interact with Kubernetes
-|   |   |-- lua    # provides utility methods to run Lua scripts
-|   |   |-- apps   # provides utility methods to manipulate Applications (e.g. start sync operation, wait for sync operation, force reconciliation)
-|   |   `-- health # provides Kubernetes resources health assessment
-|   |-- app
-|   |   |-- apis       # contains data structures that describe Application
-|   |   `-- client     # contains client that provides Application CRUD operations
-|   `-- engine     # the engine implementation
+|   |   |-- diff        # provides Kubernetes resource diffing functionality
+|   |   |-- kube        # provides utility methods to interact with Kubernetes
+|   |   |-- lua         # provides utility methods to run Lua scripts
+|   |   |-- syncunits   # provides utility methods to manipulate SyncUnits (e.g. start sync operation, wait for sync operation, force reconciliation)
+|   |   `-- health      # provides Kubernetes resources health assessment
+|   |-- syncunit
+|   |   |-- apis       # contains data structures that describe SyncUnits
+|   |   `-- client     # contains client that provides SyncUnit CRUD operations
+|   `-- engine         # the engine implementation
 ```
 
 The engine is responsible for the reconciliation of Kubernetes resources, which includes:
@@ -104,7 +104,10 @@ The manifests generation is out of scope and should be implemented by the Engine
 
 ### Engine API
 
-The engine API includes Application data structure/client and `Engine` golang interface that allows configuring reconciliation process:
+The engine API includes three main parts:
+- `Engine` golang interface
+- `SyncUnit` data structure and `SyncUnitStore` interface which provides access to the units.
+- Set of data structures which allows configuring reconciliation process.
 
 **Engine interface** - provides set of features that allows updating reconciliation settings and subscribe to engine events.
 ```golang
@@ -119,7 +122,7 @@ type Engine interface {
 	OnBeforeSync(callback func(appName string, tasks []SyncTaskInfo) ([]SyncTaskInfo, error)) Unsubscribe
 
 	// OnSyncCompleted registers callback that is executed after each sync operation.
-	OnSyncCompleted(callback func(appName string, state appv1.OperationState) error) Unsubscribe
+	OnSyncCompleted(callback func(appName string, state OperationState) error) Unsubscribe
 
 	// OnClusterCacheInitialized registers a callback that is executed when cluster cache initialization is completed.
 	OnClusterCacheInitialized(callback func(server string)) Unsubscribe
@@ -130,14 +133,33 @@ type Engine interface {
 	// OnResourceRemoved registers a callback that is executed when a cluster resource gets removed.
 	OnResourceRemoved(callback func(cluster string, key kube.ResourceKey)) Unsubscribe
 
-	// OnAppEvent registers callback that is executed on every application event.
-	OnAppEvent(callback func(app *appv1.Application, info EventInfo, message string)) Unsubscribe
+	// OnUnitEvent registers callback that is executed on every sync unit event.
+	OnUnitEvent(callback func(id string, info EventInfo, message string)) Unsubscribe
 }
 
 type Unsubscribe func()
 ```
 
-**ReconciliationSettings data structure** - holds reconciliation settings
+**SyncUnit data structure** - allows engine accessing units and update status.
+
+```golang
+type SyncUnit struct {
+	ID         string
+	Status      UnitStatus
+	Source      ManifestSource
+	Destination ManifestDestination
+	Operation   *Operation
+}
+
+type SyncUnitStore interface {
+	List() ([]SyncUnit, error)
+	Get(id string) (SyncUnit, error)
+	SetStatus(id string, status UnitStatus) (SyncUnit, error)
+	SetOperation(id string, operation *Operation) error
+}
+```
+
+**Settings data structures** - holds reconciliation settings and cluster credentials.
 ```golang
 type ReconciliationSettings struct {
 	// AppInstanceLabelKey holds label key which is automatically added to every resource managed by the application
@@ -147,10 +169,8 @@ type ReconciliationSettings struct {
 	// ResourceOverrides holds settings which customize resource diffing logic
 	ResourceOverrides map[string]appv1.ResourceOverride
 }
-```
 
-**CredentialsStore** provides access to cluster credentials
-```golang
+// provides access to cluster credentials
 type CredentialsStore interface {
 	GetCluster(ctx context.Context, name string) (*appv1.Cluster, error)
 	WatchClusters(ctx context.Context, callback func(event *ClusterEvent)) error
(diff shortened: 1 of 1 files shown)


specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
-### Public API
-
-To be documented during POC implementation.
-
+The proposed design is based on the PoC that contains draft implementation and provides the base idea of a target design. PoC is available in
+[argo-cd#fargo/engine](https://github.com/alexmt/argo-cd/tree/fargo/engine) and [flux#fargo](https://github.com/alexmt/flux/tree/fargo) 
+
+The GitOps engine API consists of three main packages:
+* The `utils` package. It consist of loosely coupled packages that implements K8S resource diffing, health assessment, etc
+* The `app` package that Application data structure definition and CRUD client interface.
```
hiddeco: What would be the (breaking) impact for Argo CD to change this into `repo` / `Repository`? I think we already discussed this multiple times, but when we detach this engine from Argo the 'application' context does not make much sense anymore, as the engine itself and Flux, unlike Argo, are focused around working with repositories.

specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
+```golang
+type CredentialsStore interface {
+	GetCluster(ctx context.Context, name string) (*appv1.Cluster, error)
+	WatchClusters(ctx context.Context, callback func(event *ClusterEvent)) error
+}
+```
+
+**ManifestGenerator** an golang interface that must be implemented by the GitOps engine consumer.
+```golang
+type ManifestResponse struct {
```
hiddeco: Something is off with the alignment here, I think it is tabs vs spaces.

specs/design-black-box.md
2opremio: +1

specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
+* The `app` package that Application data structure definition and CRUD client interface.
+* The `engine` package that leverages `utils` and uses provided set of Applications as a configuration source.
+
+```
+gitops-engine
+|-- pkg
+|   |-- utils
+|   |   |-- diff   # provides Kubernetes resource diffing functionality
+|   |   |-- kube   # provides utility methods to interact with Kubernetes
+|   |   |-- lue    # provides utility methods to run Lua scripts
```
2opremio:
```suggestion
|   |   |-- lua    # provides utility methods to run Lua scripts
```

2opremio: I think we went over this before (sorry), but why do we need Lua scripting in the core?

specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
+|   |   |-- apps   # provides utility methods to manipulate Applications (e.g. start sync operation, wait for sync operation, force reconciliation)
+|   |   `-- health # provides Kubernetes resources health assessment
+|   |-- app
+|   |   |-- apis       # contains data structures that describe Application
+|   |   `-- client     # contains client that provides Application CRUD operations
+|   `-- engine     # the engine implementation
+```
+
+The engine is responsible for reconciliation only which includes:
```
2opremio:
```suggestion
The engine is responsible for the reconciliation of Kubernetes resources, which includes:
```

specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
+- Interacting with the Kubernetes API: load and maintain the cluster state; pushing required changes to the Kubernetes API.
+- Reconciliation logic: based on provides target resource definition find which resources should be updated/deleted/created.
```
2opremio: I cannot parse this :S . typo?

specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
+- Syncing logic: determine the order in which resources should be modified; features like sync hooks, waves, etc.
+
+The manifests generation is out of scope and should be implemented by the Engine consumer.
+
+### Engine API
+
+The engine API includes Application data structure/client and `Engine` golang interface that allows configuring reconciliation process:
+
+**Engine interface** - provides set of that allows updating reconciliation settings and subscribe to engine events.
```
2opremio: set of .. functionality? features

specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
+	// SetReconciliationSettings updates reconciliation settings
+	SetReconciliationSettings(settings ReconciliationSettings)
+
+	// OnBeforeSync registers callback that is executed before each sync operation.
+	OnBeforeSync(callback func(appName string, tasks []SyncTaskInfo) ([]SyncTaskInfo, error)) Unsubscribe
+
+	// OnSyncCompleted registers callback that is executed after each sync operation.
+	OnSyncCompleted(callback func(appName string, state appv1.OperationState) error) Unsubscribe
+
+	// OnClusterCacheInitialized registers that that is executed when cluster cache initialization is completed.
```
2opremio:
```suggestion
	// OnClusterCacheInitialized registers callback that is executed when cluster cache initialization is completed.
```

specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
+	OnClusterCacheInitialized(callback func(server string)) Unsubscribe
+
+	// OnResourceUpdated registers that that is executed when cluster resource got updated.
```
2opremio:
```suggestion
	// OnResourceUpdated registers callback that is executed when cluster resource got updated.
```

specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
+	OnResourceUpdated(callback func(cluster string, un *unstructured.Unstructured)) Unsubscribe
+
+	// OnResourceRemoved registers that that is executed when cluster resource got removed.
```
2opremio:
```suggestion
	// OnResourceRemoved registers a callback that is executed when cluster resource got removed.
```

specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
+	SetReconciliationSettings(settings ReconciliationSettings)
+
+	// OnBeforeSync registers callback that is executed before each sync operation.
+	OnBeforeSync(callback func(appName string, tasks []SyncTaskInfo) ([]SyncTaskInfo, error)) Unsubscribe
+
+	// OnSyncCompleted registers callback that is executed after each sync operation.
+	OnSyncCompleted(callback func(appName string, state appv1.OperationState) error) Unsubscribe
+
+	// OnClusterCacheInitialized registers that that is executed when cluster cache initialization is completed.
+	OnClusterCacheInitialized(callback func(server string)) Unsubscribe
```
2opremio: Why is this needed?

hiddeco: Judging by the changes in the `fargo` branch of Flux, it seems to be used to kickstart the engine. https://github.com/alexmt/flux/commit/b4cc9b6a99966f75a03687a1b0599b1e8b5ef047#diff-6b48cc88be8e3f8560bbe695aa8b202fR90

specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
+```
+
+**ReconciliationSettings data structure** - holds reconciliation settings
+```golang
+type ReconciliationSettings struct {
+	// AppInstanceLabelKey holds label key which is automatically added to every resource managed by the application
+	AppInstanceLabelKey string
+	// ResourcesFilter holds settigns which allows to configure list of managed resource APIs
+	ResourcesFilter resource.ResourcesFilter
+	// ResourceOverrides holds settings which customize resource diffing logic
```
2opremio: How does this work? what should be put in the keys?

specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
+
+**CredentialsStore** provides access to cluster credentials
+```golang
+type CredentialsStore interface {
+	GetCluster(ctx context.Context, name string) (*appv1.Cluster, error)
+	WatchClusters(ctx context.Context, callback func(event *ClusterEvent)) error
+}
+```
+
+**ManifestGenerator** an golang interface that must be implemented by the GitOps engine consumer.
```
2opremio:
```suggestion
**ManifestGenerator** a Golang interface that must be implemented by the GitOps engine consumer.
```

specs/design-black-box.md
```diff
@@ -72,10 +72,121 @@ specific ConfigMaps/Secrets.
...
+```golang
+type ManifestResponse struct {
+    // Generated manifests
+    Manifests  []string
+    // Resolved Git revision
+	Revision   string
```
2opremio:
```suggestion
    Revision   string
```

specs/design-black-box.md
alexmt: Argo CD uses lua to customize health assessment.

Flux won't use it right now. It would be easy to add one more callback and remove the dependency on lua. Let me know if you want to do it please. I don't have strong opinion about it.

specs/design-black-box.md
alexmt: @hiddeco  is right. The callback is required to block `RunEngine` until cache is initialized and to make sure that Flux don't use the incomplete list of workloads.

specs/design-black-box.md
alexmt: Key is `group/kind`. Here is example of real life configuration:


```
    certmanager.k8s.io/Certificate:
      # Lua script for customizing the health status assessment
      health.lua: |
        hs = {}
        if obj.status ~= nil then
          # some logic here
        end
        hs.status = "Progressing"
        hs.message = "Waiting for certificate"
        return hs
```

Do you think it makes sense to replace string with `scheme.GroupKind` ?

specs/design-black-box.md
alexmt: fixed

specs/design-black-box.md
alexmt: I think you are right. After abstracting the manifest generation most of the Application spec is not used by the engine. Now it is much easier to come up with the Golang interface and remove the dependency on the Application.

Going through the app controller code and trying to find which methods should be exposed by the interface.  

I would like to name the interface `DeployableUnit`. This seems neutral to both Argo CD and Flux. Does the name sounds good to you? @hiddeco , @2opremio

specs/design-black-box.md
alexmt: Looks like it was too early in the morning :) fixed.

specs/design-black-box.md
alexmt: fixed

specs/design-black-box.md
hiddeco: Whenever I talk about the action Flux performs with regard to cluster operations I talk about syncing (and updating image tag references), so for me `Syncable<Something>` would make more sense to me.

jaypipes: My vote was `ManagedRepository` :)

alexmt: I guess Application is very Argo specific and Repo is very Flux specific :) 


How about `SyncUnit`? Both Argo CD and Flux uses term `syncing` so `SyncUnit` would be understandable to users from both communities.

specs/design-black-box.md
alexmt: done

specs/design-black-box.md
alexmt: done. Added comment.

specs/design-black-box.md
alexmt: Worked on draft changes in the engine and definitely looks better if we remove reference to the application. Updated design doc and removed Application from it.


I've tried using different names that include `Source` ( `SyncSource`, `TruthSource`  ) but It looks strange that Source has Source and Destination. 
e.g.

```
mySource := SyncSource{}
mySource.Source = ...
mySource.Destination = ...
```

So temporary using `SyncUnit` and totally open to change it

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
hiddeco: I am opposed to committing IDE specific files into Git. Instead `.vscode/` should be globally ignored in your `$HOME/.gitignore`, reasoning for this is about the same as [what I wrote here](https://github.com/fluxcd/flux/pull/2546#issuecomment-545838516).

alexmt: oh, that was not intentional. deleted

specs/design-black-box.md
alexmt: Replace `SyncUnit` with `Sync<Term>`. @hiddeco will setup a poll in slack to decide the best name :)

alexmt:
Merging PR to avoid blocking other docs cleanups. The final name `Sync<Term>` comes in next PR.

PR #40, opened 2020-05-20
chore: regenerate cluster cache mocks

alexmt:
PR regenerates ClusterCache mock structure.


jannfis:
I assumed the fix had to be done in ArgoCD when consuming the cache from gitops-engine. But if this is the fix, the better :)

alexmt:
@jannfis Argo CD tests failed because I forgot to regenerate the mock.

PR #49, opened 2020-06-03
refactor: add methods in errors package to quit with exit codes

darshanime:
This is a part of Streamline exit codes of CLI (https://github.com/argoproj/argo-cd/issues/3592) issue on argo-cd.  
Continued from where @jannfis [left off](https://github.com/jannfis/argo-cd/commit/4b5574ed8509098476a4bf672a5c7e8020e82b6a)

Signed-off-by: darshanime <deathbullet@gmail.com>

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -119,10 +119,10 @@ func newCmd() *cobra.Command {
 			}
 			s := settings{args[0], paths}
 			config, err := clientConfig.ClientConfig()
-			errors.CheckError(err)
+			errors.CheckErrorWithCode(err, errors.ErrorCommandSpecific)
 			if namespace == "" {
 				namespace, _, err = clientConfig.Namespace()
-				errors.CheckError(err)
+				errors.CheckErrorWithCode(err, errors.ErrorCommandSpecific)
 			}
 
 			var namespaces []string
@@ -141,10 +141,10 @@ func newCmd() *cobra.Command {
 				}),
 			)
 			gitOpsEngine := engine.NewEngine(config, clusterCache)
-			errors.CheckError(err)
+			errors.CheckErrorWithCode(err, errors.ErrorCommandSpecific)
 
 			closer, err := gitOpsEngine.Run()
-			errors.CheckError(err)
+			errors.CheckErrorWithCode(err, errors.ErrorCommandSpecific)
 
 			defer io.Close(closer)
 
@@ -162,7 +162,7 @@ func newCmd() *cobra.Command {
 				resync <- true
 			})
 			go func() {
-				errors.CheckError(http.ListenAndServe(fmt.Sprintf("0.0.0.0:%d", port), nil))
+				errors.CheckErrorWithCode(http.ListenAndServe(fmt.Sprintf("0.0.0.0:%d", port), nil), errors.ErrorCommandSpecific)
 			}()
 
 			for ; true; <-resync {
--- a/pkg/cache/cluster_test.go
+++ b/pkg/cache/cluster_test.go
@@ -18,7 +18,6 @@ import (
 	"k8s.io/client-go/dynamic/fake"
 	"k8s.io/client-go/rest"
 
-	"github.com/argoproj/gitops-engine/pkg/utils/errors"
 	"github.com/argoproj/gitops-engine/pkg/utils/kube"
 	"github.com/argoproj/gitops-engine/pkg/utils/kube/kubetest"
 )
@@ -26,7 +25,9 @@ import (
 func strToUnstructured(jsonStr string) *unstructured.Unstructured {
 	obj := make(map[string]interface{})
 	err := yaml.Unmarshal([]byte(jsonStr), &obj)
-	errors.CheckError(err)
+	if err != nil {
+		panic(err)
+	}
 	return &unstructured.Unstructured{Object: obj}
 }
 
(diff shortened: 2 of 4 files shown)


agent/main.go
```diff
@@ -119,10 +119,10 @@ func newCmd() *cobra.Command {
 			}
 			s := settings{args[0], paths}
 			config, err := clientConfig.ClientConfig()
-			errors.CheckError(err)
+			errors.CheckErrorWithCode(err, errors.ErrorCommandSpecific)
```
ash2k, line 122: I think this can be simplified if line 115 uses `RunE` field instead of `Run`. Then this function can just return the error directly and the checking logic can be centralized in `main()`.

darshanime: `CheckErrorWithCode` just exits with the appropriate exit code (instead of the default exit code) -- there is no real error checking logic, we simply exit.

agent/main.go
```diff
@@ -162,7 +162,7 @@ func newCmd() *cobra.Command {
 				resync <- true
 			})
 			go func() {
-				errors.CheckError(http.ListenAndServe(fmt.Sprintf("0.0.0.0:%d", port), nil))
+				errors.CheckErrorWithCode(http.ListenAndServe(fmt.Sprintf("0.0.0.0:%d", port), nil), errors.ErrorConnectionFailure)
```
jannfis, line 165: I think this is not a `ErrorConnectionFailure` class of error, since it creates a listener and does not try to establish a connection to the API.

pkg/cache/cluster_test.go
```diff
@@ -26,7 +26,7 @@ import (
 func strToUnstructured(jsonStr string) *unstructured.Unstructured {
 	obj := make(map[string]interface{})
 	err := yaml.Unmarshal([]byte(jsonStr), &obj)
-	errors.CheckError(err)
+	errors.CheckErrorWithCode(err, errors.ErrorCommandSpecific)
```
jannfis, line 29: I think the tests can just exit with any exit code and do not need adaption. They should just use `CheckError()` or even better, `panic()` (see below).

pkg/diff/diff_test.go
```diff
@@ -56,16 +56,16 @@ func toUnstructured(obj interface{}) (*unstructured.Unstructured, error) {
...
 }
 
 func unmarshalFile(path string) *unstructured.Unstructured {
 	data, err := ioutil.ReadFile(path)
-	errors.CheckError(err)
+	errors.CheckErrorWithCode(err, errors.ErrorCommandSpecific)
 	var un unstructured.Unstructured
 	err = json.Unmarshal(data, &un.Object)
-	errors.CheckError(err)
+	errors.CheckErrorWithCode(err, errors.ErrorCommandSpecific)
```
jannfis, line 68: Same here as with above - tests should just use `CheckError()` (or even better, I think `panic()` would be the right thing to do in tests, but we should think about this -- CC @alexmt)

pkg/utils/errors/errors.go
```diff
@@ -1,18 +1,55 @@
...
+	ErrorCommandSpecific = 1
+	// ErrorConnectionFailure is returned on connection failure to API endpoint
+	ErrorConnectionFailure = 11
+	// ErrorAPIResponse is returned on unexpected API response, i.e. authorization failure
+	ErrorAPIResponse = 12
+	// ErrorResourceDoesNotExist is returned when the requested resource does not exist
+	ErrorResourceDoesNotExist = 13
+	// ErrorGeneric is returned for generic error
+	ErrorGeneric = 20
+)
```
jannfis, line 20: We should think about whether these codes are sufficient and if the classifications are correct. The code where I left off was clearly not finished yet and not thought through to the end, so I'm unsure whether we should adapt it in this state in a verbatim way

agent/main.go
jannfis: Moving the logic of error handling to `main()` would assume that the `error` would actually hold non-generic information about the error, but I think currently most `error` variables are just generics, with nothing more than an error message produced by `fmt.Errorf()` or similar methods. But I agree, in the long run, we should have specialised `error` variables that enables to differentiate the errors directly from the `error` variable.

darshanime:
@jannfis thanks for your review. I have addressed the comments, ptal. 

Agree that the methods should return the correct error and `main()` have the centralized error handling logic. I can refactor that in a subsequent PR. In this one, I wanted to limit the scope to adding the helper methods to exit with input error codes. Let me know if it'll be better to refactor here itself.

pkg/utils/errors/errors.go
darshanime: I have used these 5 error classes for argo-cd in https://github.com/argoproj/argo-cd/pull/3700 and it seems okay. I guess we'll truly know if they are sufficient in you can review that PR 😅

agent/main.go
ash2k: I agree with the above. I just wanted to say that current approach might be less than ideal because:
- `defer`ed functions are not executed. Currently there is only one and it might not be a big deal. This might change in the future.
- Exiting on error might need something that is not accessible from a static method (e.g. a field or method on some object). E.g. the program may want to send an event to a metric/error aggregation system. Having multiple places that do this (vs one in the `main()`) might be more verbose.
- Having 1 entry point and 1 exit point is easier to follow in general, but this is subjective.

agent/main.go
```diff
@@ -141,10 +141,10 @@ func newCmd() *cobra.Command {
 				}),
 			)
 			gitOpsEngine := engine.NewEngine(config, clusterCache)
-			errors.CheckError(err)
+			errors.CheckErrorWithCode(err, errors.ErrorCommandSpecific)
```
ash2k, line 144: Was looking at the code and noticed that this check is not needed. Trying to use the engine...

PR #51, opened 2020-06-03
fix: race condition (#50)

d-kuro:
refs: https://github.com/argoproj/gitops-engine/issues/50

If you import the code from https://github.com/argoproj/gitops-engine/pull/52, the race condition will be gone.

```shell
# refs: https://github.com/argoproj/gitops-engine/pull/52
$ git status
On branch feature/fix-race-condition
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   pkg/sync/sync_context_test.go
        modified:   pkg/utils/kube/kubetest/mock.go

no changes added to commit (use "git add" and/or "git commit -a")

$ go test ./... --race
?       github.com/argoproj/gitops-engine       [no test files]
?       github.com/argoproj/gitops-engine/agent [no test files]
ok      github.com/argoproj/gitops-engine/pkg/cache     (cached)
?       github.com/argoproj/gitops-engine/pkg/cache/mocks       [no test files]
ok      github.com/argoproj/gitops-engine/pkg/diff      (cached)
?       github.com/argoproj/gitops-engine/pkg/engine    [no test files]
ok      github.com/argoproj/gitops-engine/pkg/health    (cached)
ok      github.com/argoproj/gitops-engine/pkg/sync      (cached)
ok      github.com/argoproj/gitops-engine/pkg/sync/common       (cached)
ok      github.com/argoproj/gitops-engine/pkg/sync/hook (cached)
ok      github.com/argoproj/gitops-engine/pkg/sync/hook/helm    (cached)
ok      github.com/argoproj/gitops-engine/pkg/sync/ignore       (cached)
ok      github.com/argoproj/gitops-engine/pkg/sync/resource     (cached)
ok      github.com/argoproj/gitops-engine/pkg/sync/syncwaves    (cached)
?       github.com/argoproj/gitops-engine/pkg/utils/errors      [no test files]
ok      github.com/argoproj/gitops-engine/pkg/utils/exec        (cached)
?       github.com/argoproj/gitops-engine/pkg/utils/io  [no test files]
?       github.com/argoproj/gitops-engine/pkg/utils/json        [no test files]
ok      github.com/argoproj/gitops-engine/pkg/utils/kube        (cached)
?       github.com/argoproj/gitops-engine/pkg/utils/kube/kubetest       [no test files]
?       github.com/argoproj/gitops-engine/pkg/utils/testing     [no test files]
?       github.com/argoproj/gitops-engine/pkg/utils/text        [no test files]
ok      github.com/argoproj/gitops-engine/pkg/utils/tracing     (cached)
```

Among many functions, using `sync.Mutex` to secure a lock was complex and difficult.
I took the approach of using `sync.Map`.
I don't think this is the best response, but I think it's good that race conditions are gone.
If you have any comments, please feel free to do so!

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -3,7 +3,6 @@ package cache
 import (
 	"context"
 	"fmt"
-	"runtime/debug"
 	"sort"
 	"strings"
 	"sync"
@@ -412,14 +411,8 @@ func runSynced(lock sync.Locker, action func() error) error {
 }
 
 func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo, info *apiMeta, resClient dynamic.ResourceInterface, ns string) {
-	kube.RetryUntilSucceed(func() (err error) {
-		defer func() {
-			if r := recover(); r != nil {
-				err = fmt.Errorf("Recovered from panic: %+v\n%s", r, debug.Stack())
-			}
-		}()
-
-		err = runSynced(&c.lock, func() error {
+	kube.RetryUntilSucceed(func() error {
+		err := runSynced(&c.lock, func() error {
 			if info.resourceVersion == "" {
 				listPager := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
 					res, err := list(resClient, opts)
@@ -429,7 +422,7 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
 					return res, err
 				})
 				var items []unstructured.Unstructured
-				err = listPager.EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
+				err := listPager.EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
 					if un, ok := obj.(*unstructured.Unstructured); !ok {
 						return fmt.Errorf("object %s/%s has an unexpected type", un.GroupVersionKind().String(), un.GetName())
 					} else {
@@ -521,7 +514,7 @@ func (c *clusterCache) processApi(client dynamic.Interface, api kube.APIResource
 	return nil
 }
 
-func (c *clusterCache) sync() (err error) {
+func (c *clusterCache) sync() error {
 	c.log.Info("Start syncing cluster")
 
 	for i := range c.apisMeta {
@@ -574,7 +567,7 @@ func (c *clusterCache) sync() (err error) {
 				return res, err
 			})
 
-			err = listPager.EachListItem(context.Background(), metav1.ListOptions{}, func(obj runtime.Object) error {
+			err := listPager.EachListItem(context.Background(), metav1.ListOptions{}, func(obj runtime.Object) error {
 				if un, ok := obj.(*unstructured.Unstructured); !ok {
 					return fmt.Errorf("object %s/%s has an unexpected type", un.GroupVersionKind().String(), un.GetName())
 				} else {
@@ -631,6 +624,7 @@ func (c *clusterCache) GetNamespaceTopLevelResources(namespace string) map[kube.
 	defer c.lock.RUnlock()
 	resources := make(map[kube.ResourceKey]*Resource)
 	for _, res := range c.nsIndex[namespace] {
+		res := res
 		if len(res.OwnerRefs) == 0 {
 			resources[res.ResourceKey()] = res
 		}
@@ -647,6 +641,7 @@ func (c *clusterCache) IterateHierarchy(key kube.ResourceKey, action func(resour
 		action(res, nsNodes)
 		childrenByUID := make(map[types.UID][]*Resource)
 		for _, child := range nsNodes {
+			child := child
 			if res.isParentOf(child) {
 				childrenByUID[child.Ref.UID] = append(childrenByUID[child.Ref.UID], child)
 			}
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -60,22 +60,23 @@ func list(resClient dynamic.ResourceInterface, opts metav1.ListOptions) (*unstru
 }
 
 type apiMeta struct {
-	lock            sync.Mutex
 	namespaced      bool
-	resourceVersion string
+	resourceVersion atomic.Value
 	watchCancel     context.CancelFunc
 }
 
-func (a *apiMeta) GetResourceVersion() string {
-	a.lock.Lock()
-	defer a.lock.Unlock()
-	return a.resourceVersion
+func (a *apiMeta) LoadResourceVersion() (string, error) {
+	v := a.resourceVersion.Load()
+	version, ok := v.(string)
+	if ok {
+		return version, nil
+	}
+
+	return "", fmt.Errorf("stored type is invalid in resourceVersion")
 }
 
-func (a *apiMeta) UpdateResourceVersion(new string) {
-	a.lock.Lock()
-	defer a.lock.Unlock()
-	a.resourceVersion = new
+func (a *apiMeta) StoreResourceVersion(resourceVersion string) {
+	a.resourceVersion.Store(resourceVersion)
 }
 
 // ClusterInfo holds cluster cache stats
@@ -376,7 +377,7 @@ func (c *clusterCache) replaceResourceCache(gk schema.GroupKind, resourceVersion
 		if err != nil {
 			return err
 		}
-		info.UpdateResourceVersion(resourceVersion)
+		info.StoreResourceVersion(resourceVersion)
 	}
 
 	return nil
@@ -551,11 +552,16 @@ func runSynced(action func() error) error {
 func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo, info *apiMeta, resClient dynamic.ResourceInterface, ns string) {
 	kube.RetryUntilSucceed(func() error {
 		err := runSynced(func() error {
-			if info.GetResourceVersion() == "" {
+			version, err := info.LoadResourceVersion()
+			if err != nil {
+				return err
+			}
+
+			if version == "" {
 				listPager := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
 					res, err := list(resClient, opts)
 					if err == nil {
-						info.UpdateResourceVersion(res.GetResourceVersion())
+						info.StoreResourceVersion(res.GetResourceVersion())
 					}
 					return res, err
 				})
@@ -571,7 +577,13 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
 				if err != nil {
 					return fmt.Errorf("failed to load initial state of resource %s: %v", api.GroupKind.String(), err)
 				}
-				if err := c.replaceResourceCache(api.GroupKind, info.GetResourceVersion(), items, ns); err != nil {
+
+				version, err := info.LoadResourceVersion()
+				if err != nil {
+					return err
+				}
+
+				if err := c.replaceResourceCache(api.GroupKind, version, items, ns); err != nil {
 					return err
 				}
 			}
@@ -582,7 +594,12 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
 			return err
 		}
 
-		w, err := resClient.Watch(metav1.ListOptions{ResourceVersion: info.GetResourceVersion()})
+		version, err := info.LoadResourceVersion()
+		if err != nil {
+			return err
+		}
+
+		w, err := resClient.Watch(metav1.ListOptions{ResourceVersion: version})
 		if errors.IsNotFound(err) {
 			if err := c.stopWatching(api.GroupKind, ns); err != nil {
 				return err
@@ -591,7 +608,7 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
 		}
 
 		if errors.IsGone(err) {
-			info.UpdateResourceVersion("")
+			info.StoreResourceVersion("")
 			c.log.Warnf("Resource version of %s is too old", api.GroupKind)
 		}
 
@@ -607,7 +624,7 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
 			case event, ok := <-w.ResultChan():
 				if ok {
 					obj := event.Object.(*unstructured.Unstructured)
-					info.UpdateResourceVersion(obj.GetResourceVersion())
+					info.StoreResourceVersion(obj.GetResourceVersion())
 					c.processEvent(event.Type, obj)
 					if kube.IsCRD(obj) {
 						if event.Type == watch.Deleted {
@@ -710,7 +727,7 @@ func (c *clusterCache) sync() error {
 			listPager := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
 				res, err := list(resClient, opts)
 				if err == nil {
-					info.UpdateResourceVersion(res.GetResourceVersion())
+					info.StoreResourceVersion(res.GetResourceVersion())
 				}
 				return res, err
 			})
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -124,7 +124,7 @@ type ClusterCache interface {
 func NewClusterCache(config *rest.Config, opts ...UpdateSettingsFunc) *clusterCache {
 	cache := &clusterCache{
 		settings:                Settings{ResourceHealthOverride: &noopSettings{}, ResourcesFilter: &noopSettings{}},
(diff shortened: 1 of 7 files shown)


ash2k:
Hey, you beat me to it! I was also working on fixing these race conditions. I think your branch does more and does it better as I haven't tried to simplify things as much - I'm not that familiar with the codebase. You can see what I've done here if you are interested https://github.com/argoproj/gitops-engine/compare/master...ash2k:ash2k/fixes

pkg/sync/sync_context.go
```diff
@@ -753,60 +759,71 @@ func (sc *syncContext) runTasks(tasks syncTasks, dryRun bool) runState {
...
+							runStateMutex.Lock()
 							runState = failed
+							runStateMutex.Unlock()
 							sc.setResourceResult(t, "", common.OperationError, fmt.Sprintf("failed to delete resource: %v", err))
 						}
 					} else {
 						// if there is anything that needs deleting, we are at best now in pending and
 						// want to return and wait for sync to be invoked again
+						runStateMutex.Lock()
 						runState = pending
```
ash2k, line 771: I think apart from the data race that you are fixing here, there is also a logic error. One goroutine may set `runState` to `failed` but then another one to `pending` or the other way around. This non-deterministic behavior is probably not right, or is it? Please see https://github.com/argoproj/gitops-engine/commit/1e0e79249b8b390d0736f753a7d2d490c0a90c41 for a different approach.

pkg/sync/sync_context.go
```diff
@@ -651,7 +651,7 @@ func (sc *syncContext) Terminate() {
 				sc.setResourceResult(task, "", common.OperationFailed, fmt.Sprintf("Failed to delete: %v", err))
 				terminateSuccessful = false
 			} else {
-				sc.setResourceResult(task, "", common.OperationSucceeded, fmt.Sprintf("Deleted"))
```
alexmt, line 654: 👍

pkg/sync/sync_context.go
```diff
@@ -721,14 +723,17 @@ func (sc *syncContext) runTasks(tasks syncTasks, dryRun bool) runState {
 	{
 		var wg sync.WaitGroup
 		for _, task := range pruneTasks {
+			task := task
```
alexmt, line 726: 👍

d-kuro:
@ash2k @alexmt 
Thank you for your comment!
I'm going to work on the code fix that was pointed out to me over the weekend.

ash2k:
@d-kuro 👍 

I suggest to always run `go test -race`. ArgoCD's tests too!

alexmt:
v1.6 testing looks good so far. We just want to run it internally a few more days and create 1.6 GA release. As soon as 1.6 is out I will test this PR's performance.

pkg/sync/sync_context.go
ash2k: This line is not needed as `task` is passed to the closure (line 737/742) and it uses `t` instead.

pkg/sync/sync_context.go
```diff
@@ -743,6 +748,7 @@ func (sc *syncContext) runTasks(tasks syncTasks, dryRun bool) runState {
 	if runState == successful && createTasks.Any(func(t *syncTask) bool { return t.needsDeleting() }) {
 		var wg sync.WaitGroup
 		for _, task := range createTasks.Filter(func(t *syncTask) bool { return t.needsDeleting() }) {
+			task := task
```
ash2k, line 751: This line is not needed, same as above.

pkg/sync/sync_context.go
```diff
@@ -707,10 +707,12 @@ func (sc *syncContext) runTasks(tasks syncTasks, dryRun bool) runState {
 	sc.log.WithFields(log.Fields{"numTasks": len(tasks), "dryRun": dryRun}).Debug("running tasks")
 
 	runState := successful
+	var runStateMutex sync.Mutex
 	var createTasks syncTasks
 	var pruneTasks syncTasks
 
 	for _, task := range tasks {
+		task := task
```
ash2k, line 715: I wonder why is this needed?

pkg/cache/cluster.go
```diff
@@ -32,11 +31,24 @@ const (
 )
 
 type apiMeta struct {
+	lock            sync.Mutex
 	namespaced      bool
 	resourceVersion string
```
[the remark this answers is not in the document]
d-kuro: Fixed
https://github.com/argoproj/gitops-engine/pull/51/commits/8c8713d18297a41d000a25f580d324d753b898b8

pkg/cache/cluster.go
```diff
@@ -376,31 +515,23 @@ func (c *clusterCache) startMissingWatches() error {
 	return nil
 }
 
-func runSynced(lock sync.Locker, action func() error) error {
-	lock.Lock()
-	defer lock.Unlock()
+func runSynced(action func() error) error {
```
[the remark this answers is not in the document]
d-kuro: Thanks!
https://github.com/argoproj/gitops-engine/pull/51/commits/29224aaf138d6775fb55a25ce8a4fbfb8527119a

pkg/sync/sync_context.go
d-kuro: I cherry-picked from your commit.
Thank you!

https://github.com/argoproj/gitops-engine/pull/51/commits/6a762b3b226aa4aa8be9d643dd7538633d4581e5
https://github.com/argoproj/gitops-engine/pull/51/commits/c5d445f642a1283a09aa0c0056274e1e6fe17997

pkg/sync/sync_context.go
d-kuro: I think that's been fixed with this commit.
https://github.com/argoproj/gitops-engine/ pull/51/commits/c5d445f642a1283a09aa0c 0056274e1e6fe17997

pkg/sync/sync_context.go
d-kuro: I understand.
Thank you comments.

I think that's been fixed with this commit.
https://github.com/argoproj/gitops-engine/ pull/51/commits/c5d445f642a1283a09aa0c 0056274e1e6fe17997

pkg/sync/sync_context.go
d-kuro: I think that's been fixed with this commit.
https://github.com/argoproj/gitops-engine/pull/51/commits/c5d445f642a1283a09aa0c0056274e1e6fe17997

pkg/cache/cluster.go
```diff
@@ -62,10 +61,24 @@ func list(resClient dynamic.ResourceInterface, opts metav1.ListOptions) (*unstru
...
 }
 
+func (a *apiMeta) LoadResourceVersion() (string, error) {
+	v := a.resourceVersion.Load()
+	version, ok := v.(string)
+	if ok {
+		return version, nil
+	}
+
+	return "", fmt.Errorf("stored type is invalid in resourceVersion")
```
ash2k, line 75: This cannot happen and if it does it should be ok to panic because... it should never happen?
I'd do this instead

```go
str, _ := a.resourceVersion.Load().(string)
return str, nil
```
This will not error out and it also preserves current semantics where the field is an empty string if it was never set. It's also nicer to not need to check the error in every place where this method is called.

pkg/cache/cluster.go
```diff
@@ -141,19 +154,56 @@ func NewClusterCache(config *rest.Config, opts ...UpdateSettingsFunc) *clusterCa
...
 
+type Resources struct {
+	sync.Map
+}
+
+func (r *Resources) StoreResources(key kube.ResourceKey, value *Resource) {
+	r.Store(key, value)
+}
+
+func (r *Resources) LoadResources(key kube.ResourceKey) (*Resource, error) {
```
ash2k, line 165: Same as above - panicking should be ok if something unexpected is in this private field. And the error is not checked in a few places below. Why return it then? :)

pkg/cache/cluster.go
```diff
@@ -167,6 +217,67 @@ type clusterCache struct {
 	eventHandlers               map[uint64]OnEventHandler
 }
 
+func (c *clusterCache) StoreApisMeta(key schema.GroupKind, value *apiMeta) {
+	c.apisMeta.Store(key, value)
+}
+
+func (c *clusterCache) LoadApisMeta(key schema.GroupKind) (*apiMeta, error) {
```
ash2k, line 224: Same as above - panicking should be ok if something unexpected is in this private field.

pkg/cache/cluster.go
```diff
@@ -167,6 +217,67 @@ type clusterCache struct {
...
+		return true
+	})
+	return cnt
+}
+
+func (c *clusterCache) StoreNsIndex(key string, value *Resources) {
+	c.nsIndex.Store(key, value)
+}
+
+func (c *clusterCache) LoadNsIndex(key string) (*Resources, error) {
```
ash2k, line 259: Same as above - panicking should be ok if something unexpected is in this private field.

pkg/cache/cluster.go
```diff
@@ -223,9 +334,12 @@ func (c *clusterCache) GetAPIGroups() []metav1.APIGroup {
...
 
-func (c *clusterCache) replaceResourceCache(gk schema.GroupKind, resourceVersion string, objs []unstructured.Unstructured, ns string) {
-	info, ok := c.apisMeta[gk]
-	if ok {
+func (c *clusterCache) replaceResourceCache(gk schema.GroupKind, resourceVersion string, objs []unstructured.Unstructured, ns string) error {
+	info, err := c.LoadApisMeta(gk)
+	if err != nil {
+		return err
+	}
+	if info != nil {
```
ash2k, line 342: This `if` can be inverted to avoid so much indentation. https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

pkg/cache/cluster.go
```diff
@@ -611,8 +760,6 @@ func (c *clusterCache) EnsureSynced() error {
 		return c.syncError
 	}
 
-	c.lock.Lock()
-	defer c.lock.Unlock()
 	// before doing any work, check once again now that we have the lock, to see if it got
 	// synced between the first check and now
```
ash2k, line 764: This comment and code above can be deleted now.

pkg/cache/cluster.go
```diff
@@ -626,31 +773,50 @@ func (c *clusterCache) EnsureSynced() error {
...
 // GetNamespaceTopLevelResources returns top level resources in the specified namespace
-func (c *clusterCache) GetNamespaceTopLevelResources(namespace string) map[kube.ResourceKey]*Resource {
-	c.lock.RLock()
-	defer c.lock.RUnlock()
-	resources := make(map[kube.ResourceKey]*Resource)
-	for _, res := range c.nsIndex[namespace] {
-		if len(res.OwnerRefs) == 0 {
-			resources[res.ResourceKey()] = res
+func (c *clusterCache) GetNamespaceTopLevelResources(namespace string) *Resources {
+	res, _ := c.LoadNsIndex(namespace)
```
ash2k, line 777: No error check here too. That's why I'm suggesting to just let it panic if the field contents is corrupted. The next line will panic anyway.

d-kuro:
@ash2k 

> I think it might be not the best idea to use sync.Map because it makes it harder to write code:

I understand.
I too think the `Range()` of the sync map is not good and should be improved.
Initially, we decided to use sync.Map because of the complexity of locking on the map, but the Now that we've cleaned up the code to access it, we can use the mutex and map access approach with I think we can switch it up.

I'm going to try to change it back from sync map to map once.
Thanks!

pkg/cache/cluster.go
```diff
@@ -779,25 +969,35 @@ func (c *clusterCache) onNodeUpdated(oldRes *Resource, un *unstructured.Unstruct
...
-	existing, ok := c.resources[key]
-	if ok {
-		delete(c.resources, key)
-		ns, ok := c.nsIndex[key.Namespace]
-		if ok {
-			delete(ns, key)
-			if len(ns) == 0 {
-				delete(c.nsIndex, key.Namespace)
+func (c *clusterCache) onNodeRemoved(key kube.ResourceKey) error {
+	existing, err := c.resources.LoadResources(key)
```
alexmt, line 978: I agree with @ash2k regarding `sync.Map` usage.

I think we still need to have a lock here to make sure `resources` map and namespace index are synchronized.

pkg/cache/cluster.go
alexmt, line 259: I think `LoadNsIndex` is a private method. Please rename to `loadNsIndex`.

pkg/cache/cluster.go
```diff
@@ -167,6 +217,67 @@ type clusterCache struct {
...
+func (c *clusterCache) ApisMetaLength() int {
+	var cnt int
+	c.apisMeta.Range(func(key, value interface{}) bool {
+		cnt++
+		return true
+	})
+	return cnt
+}
+
+func (c *clusterCache) StoreNsIndex(key string, value *Resources) {
```
alexmt, line 255: Same as `LoadNsIndex`. Please rename to `storeNsIndex`.

alexmt:
Hello @d-kuro . I'm very interested in completing and merging this PR. Please, let me know if you have time to continue working on it. 
I can help. E.g. I can send PR to your fork that addresses `sync.Map` related comments. let me know

d-kuro:
@alexmt 
Sorry for the late reply.

I'm going to try migrating from SyncMap to map, as I wrote in my comment below.
https://github.com/argoproj/gitops-engine/pull/51#issuecomment-646390088

I may ask you to follow up if you try and get stuck, but I'll try to do it myself for now.
Thanks for your concern!

d-kuro:
@ash2k cc: @alexmt 

I'm trying to work on migration from `sync.Map` to `map`.
but it suffers from the race condition of map range access and writing.

> fatal error: concurrent map iteration and map write

Using RLock for map range access and Lock for write is simple but for example the following code.
https://github.com/argoproj/gitops-engine/blob/master/pkg/cache/cluster.go#L241-L249

Deadlocked this function, because we are deleting the map in range access that requires a read lock.
Is there a good way to write this without using `SyncMap`?

```go
// c.resources is "map[kube.ResourceKey]*Resource"

// Need RWMutex.RLock()
for key := range c.resources {
	if key.Kind != gk.Kind || key.Group != gk.Group || ns != "" && key.Namespace != ns {
		continue
	}

	if _, ok := objByKey[key]; !ok {
		// Need RWMutex.Lock()
		// "delete(c.resources, key)" in a function
		c.onNodeRemoved(key)
	}
}
```

ash2k:
@d-kuro Perhaps `replaceResourceCache()` can be left as is but then `watchEvents()` which calls it can grab the write lock before calling it?

I'd like to suggest to split this PR as it's getting quite big. Perhaps make all the changes to `pkg/cache/cluster.go` in a separate PR?

I've also looked at the `clusterCache` more closely and I'm wondering why was it written using list+watch+manual cache (`resources` field) and not using [informers](https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go)? It might be a good idea to refactor `clusterCache` to use informers instead - they support notifications, indexes and are thoroughly tested by kubernetes inself. WDYT?

alexmt:
Thanks a lot for continue working on it  @d-kuro !


I've looked only more time where `replaceResourceCache` is used. The method is executed in two places and both have write lock:

https://github.com/argoproj/gitops-engine/blob/ce9616ad10da6d05c78ecceba23ae256ccb0c0f2/pkg/cache/cluster.go#L443

https://github.com/argoproj/gitops-engine/blob/ce9616ad10da6d05c78ecceba23ae256ccb0c0f2/pkg/cache/cluster.go#L371


I suspect race condition happens in tests only: [TestNamespaceModeReplace](https://github.com/argoproj/gitops-engine/blob/7d3da9f16e353692824a6e83fc31de113b6a9251/pkg/cache/cluster_test.go#L378). I hope I'm not missing anything. If this is the case we just need to use lock in `TestNamespaceModeReplace` test.

alexmt:
@ash2k we could not use informers framework because of informers cache the whole resource manifest. We need only resources reference and parent links. Switching to informers would increase memory consumption.

ash2k:
@alexmt 

> we could not use informers framework because of informers cache the whole resource manifest.

This is the default behavior, but it is configurable. If only certain fields are required, a custom Go type can be crafted to use for any watched resource type. That way only the needed fields are unmarshaled and nothing else. For example, there is a whole package just for [working with metadata of objects](https://github.com/kubernetes/client-go/tree/master/metadata). Tuned client and informer are available there.

However, memory consumption will still be higher in the above case vs how things currently work because e.g. all annotations will be stored in the informer's cache. Thank you for answering the question :)

alexmt:
Changes of this PR got split into several other PRs and race detection is now enabled in the project. @d-kuro , @ash2k thanks a lot for driving it!

alswl:
`c.lock` and `c.listSemaphore` had dead lock situation in big clusters. `c.lock` holds too big scope in `sync()` function, we should remove `c.lock`, using `sync.Map`.

PR #55, opened 2020-06-08
feat: support limiting number of concurrent k8s list queries

alexmt:
PR introduces an ability to limit the number of k8s list queries that helps to avoid memory spikes during controller initialization and resolves https://github.com/argoproj/argo-cd/issues/3241

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -40,6 +40,7 @@ var (
 )
 
 // SetMaxConcurrentList set maximum number of concurrent K8S list calls.
+// If set to 0 then no limit is enforced.
 // Note: method is not thread safe. Use it during initialization before executing ClusterCache.EnsureSynced method.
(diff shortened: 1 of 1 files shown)


pkg/cache/cluster.go
```diff
@@ -31,6 +32,33 @@ const (
...
 )
 
+var (
+	// listSemaphore is used to limit the number of concurrent k8s list queries.
+	// Global limit is required to avoid memory spikes during cache initialization.
+	// The default limit of 50 is chosen based on experiments.
+	listSemaphore = semaphore.NewWeighted(50)
+)
+
+// SetMaxConcurrentList set maximum number of concurrent K8S list calls.
```
jannfis, line 42: I think it should be documented that if `val` is less or equal to `0`, limit is disabled.

alexmt: done.

PR #58, opened 2020-06-11
refactor: use UpdateSettingsFunc type

ash2k:
A small cleanup.


pkg/cache/cluster.go
```diff
@@ -122,7 +122,7 @@ type ClusterCache interface {
 }
 
 // NewClusterCache creates new instance of cluster cache
-func NewClusterCache(config *rest.Config, opts ...func(cache *clusterCache)) *clusterCache {
```
alexmt, line 125: 👍

PR #60, opened 2020-06-20
chore: Add IgnoreResourceStatusField field to DiffOptions

darshanime:
Can be used to ignore `/status` of the objects. 
See https://github.com/argoproj/argo-cd/pull/3754

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -31,7 +31,8 @@ import (
 // Holds diffing settings
 type DiffOptions struct {
 	// If set to true then differences caused by aggregated roles in RBAC resources are ignored.
-	IgnoreAggregatedRoles bool `json:"ignoreAggregatedRoles,omitempty"`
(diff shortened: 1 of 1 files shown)


darshanime:
I thought about it, but then thought not all consumers of this library would want to abide by that classification. If it's a string field, they can use it in a way that fits their semantics. 

What do you think?

jannfis:
Hm yes, that's a good point. 

But now I'm also thinking about whether we should rather move the whole functionality into gitops-engine, instead of making it an argo-cd feature with a "dead" switch in gitops-engine. Or should we move the diff options (and its current features) to argo-cd?

alexmt:
Instead of adding an unused field, I would suggest adding data structure that inherits `DiffOptions` from gitops engine and adds `IgnoreResourceStatusField`.

darshanime:
This can be done in `argo-cd`, can I close this PR in that case?

alexmt:
sure @darshanime . lets clone this PR and implement changes in Argo CD

darshanime:
cool, thanks for comments, closing

PR #61, opened 2020-06-20
fix: Check for err != nil before Fatal

darshanime:
`CheckError` did not check if errors are `nil` before calling `Fatal`. 
Fixes regression introduced in 16575f18343d678f6beffde648bc6d54bc4b8a63


jannfis:
LGTM, thanks for fixing it.

PR #63, opened 2020-06-23
refactor: Drop github.com/grpc-ecosystem/grpc-gateway dependency

ash2k:
gitops-engine does not use the `JSONMarshaler` type and pulling in a dependency for it is not great. Let's remove it please.

I'm trying to optimize the amount of stuff my project, that consumes the gitops-engine, needs to pull down and compile to work.

### DIFF
--- a/pkg/utils/json/json.go
+++ b/pkg/utils/json/json.go
@@ -1,40 +1,5 @@
 package json
 
-import (
-	"encoding/json"
-	"io"
-
-	gwruntime "github.com/grpc-ecosystem/grpc-gateway/runtime"
-)
-
-// JSONMarshaler is a type which satisfies the grpc-gateway Marshaler interface
-type JSONMarshaler struct{}
-
-// ContentType implements gwruntime.Marshaler.
-func (j *JSONMarshaler) ContentType() string {
-	return "application/json"
-}
-
-// Marshal implements gwruntime.Marshaler.
-func (j *JSONMarshaler) Marshal(v interface{}) ([]byte, error) {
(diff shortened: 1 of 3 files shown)


jannfis:
This broke build of argo-cd, because the JSON marshaller is used in `server/server.go`. We need to move it into argo-cd codebase.

alexmt:
Thank you for heads up @jannfis . I will work on PR that upgrades gitops engine version and moves marshaler into Argo CD code.

pkg/utils/json/json.go
```diff
@@ -79,12 +44,3 @@ func removeListFields(config, live []interface{}) []interface{} {
 	}
 	return result
 }
-
-// MustMarshal is a convenience function to marshal an object successfully or panic
-func MustMarshal(v interface{}) []byte {
```
jannfis, line 84: I think this method should have stayed here.

ash2k: I removed it because it is not used. I personally see gitops-engine as a library and as such I think it'd be best for it to have a small clean API with as little extra stuff exposed as possible. Reducing the exposed API surface has the following benefits:
- It gives maintainers the biggest flexibility to change how things work without breaking users. If a method/function/type is not exposed then changing it will not break any users.
- It makes the library easy to use.

jannfis: I agree, the less methods we expose through the API, the better. The JSON utils were probably never supposed to be consumed by consumers anyway, so better to break existing consumers now while we're still pre-v1.0 than later. Thanks for your response. :+1:

PR #64, opened 2020-06-24
fix: Do not perform client side conversion of K8s resources

jannfis:
Do not perform client side conversion of resources and instead retrieve the converted resource from K8s.

ArgoCD linked to gitops-engine with this change passes end-to-end and unit tests.

Fixes https://github.com/argoproj/argo-cd/issues/3670


jannfis:
Deprecated

PR #80, opened 2020-07-07
feat: Add func GetManagedObject

cyrus-mc:
Retrieve object from cache based on `ResourceKey`. Cache manifest if it isn't already cached. 

This is required in support of https://github.com/argoproj/argo-cd/issues/3854 and https://github.com/argoproj/argo-cd/pull/3863.


cyrus-mc:
Need to base off previous commit as latest changes break argoCD.

PR #82, opened 2020-07-08
fix: diff should perform server side like apply during resource diffing

alexmt:
PR closes https://github.com/argoproj/argo-cd/issues/2865 . 

The only way to properly test built-in Kubernetes types without using K8S server is to execute server-side logic on the client. The PR implements server side k8s apply logic: uses built-in k8s type structure, apply default values and use strategic merge patch .

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -26,6 +26,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	jsonutil "github.com/argoproj/gitops-engine/pkg/utils/json"
+	kubescheme "github.com/argoproj/gitops-engine/pkg/utils/kube/scheme"
 )
 
 // Holds diffing settings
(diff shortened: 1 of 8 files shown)


pkg/diff/diff.go
```diff
@@ -118,22 +119,43 @@ func TwoWayDiff(config, live *unstructured.Unstructured) (*DiffResult, error) {
...
 	}
 
-	liveTemplate, ok, err := unstructured.NestedSlice(live.Object, "spec", "volumeClaimTemplates")
-	if !ok || err != nil {
-		return
+	err = json.Unmarshal(predictedLiveBytes, &predictedLive)
+	if err != nil {
+		return nil, nil, err
+	}
+	kubescheme.Scheme.Default(predictedLive)
```
ash2k, line 138: gitops-engine uses libraries from Kubernetes version X, but user might be using Kubernetes version Y. This creates issues with this approach:
- I don't think defaults are guaranteed to be the same across Kubernetes versions. Version X and Y might default the same field to different values.
-  A resource kind of the same API version in newer version Y might have new fields or might have dropped fields and hence defaulting will work different in those versions.
- Defaulting sometimes depends on feature gates, which can be enabled via api server configuration. I don't know how common this is, but I ran into this at least once. This is probably the most annoying thing from this list.

Also:
- What about custom resources with their defaulting?
- What about custom API servers and their defaulting? (e.g. Service Catalog)

alexmt: I agree, @ash2k . Looks like there is no perfect way to compare resources on the client-side. This PR is an intermediate solution. It supposed to "stop the bleeding" and improves the diffing of built-in Kubernetes such as Deployment.

The next step is to use K8S server-side apply dry run to get an accurate diff.

jessesuen: We understand that client side diffing will never be perfect -- I had brought up some of the same points about client version potentially cause differences depending on Kubernetes server versions. Some of your other points are new to me (e.g. defaulting controlled by feature gates or custom API servers).

That said, I think the approach being taken here is at least better than what we have now, which is that we incorrectly reporting diffs for some very common scenarios. I believe in the vast majority of cases, the defaulting of unspecified fields does not cause a problem, since argo-cd should not consider the **extra** fields to contribute to the "diff". But in some corner cases, the manner in which the controller is defaulting fields (e.g. StatefulSets) is causing problems which can only be solved by defaulting.

> What about custom resources with their defaulting?

This was also raised, and an area of future improvement. We would have to understand OpenAPI schema in the CRD and default based on the spec.

Given that the only way to get a perfect diff is through the API server, another approach was considered where we relied on the server to perform the diff (exactly like kubectl diff). However this would dramatically increase the burden on the API server. Also, the Argo CD CLI also has the ability perform a local diff, and the feature would be lost with this approach.

I think we server-side diff may eventual be direction we need to go, but in order to do so, we'd need to implement some additional caching (e.g. keyed on `metadata.generation`), in order to prevent the additional load against the K8s API caused by the frequency of application reconciliation.

jessesuen: > since argo-cd should not consider the extra fields to contribute to the "diff".

@alexmt can you confirm that this still true with this approach?

alexmt: @jessesuen , that is right. This approach solves the `StatefulSets` edge case.

PR #84, opened 2020-07-09
feat: improve sync operation messages

alexmt:
PR improves the sync state messages that provide more visibility into why the syncing process takes time. 

After PR is merged the sync state will have a detailed message instead of `one or more tasks are running`. Examples:

* waiting for healthy state of /Pod/my-pod and 2 more resources
* waiting for completion of hook /Pod/my-pod

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -243,6 +243,27 @@ type syncContext struct {
 	lock sync.Mutex
 }
 
+func (sc *syncContext) setRunningPhase(tasks []*syncTask, isPendingDeletion bool) {
+	if len(tasks) > 0 {
+		firstTask := tasks[0]
+		waitingFor := "completion of hook"
+		andMore := "hooks"
+		if !firstTask.isHook() {
+			waitingFor = "healthy state of"
+			andMore = "resources"
+		}
+		if isPendingDeletion {
+			waitingFor = "deletion of"
+		}
+		message := fmt.Sprintf("waiting for %s %s/%s/%s",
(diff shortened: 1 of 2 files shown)


alexmt:
Related to https://github.com/argoproj/argo-cd/issues/3663

pkg/sync/sync_context.go
```diff
@@ -243,6 +243,27 @@ type syncContext struct {
...
+			andMore = "resources"
+		}
+		if isPendingDeletion {
+			waitingFor = "deletion of"
+		}
+		message := fmt.Sprintf("waiting for %s %s/%s/%s",
+			waitingFor, firstTask.group(), firstTask.kind(), firstTask.name())
+		if moreTasks := len(tasks) - 1; moreTasks > 0 {
+			message = fmt.Sprintf("%s and %d more %s", message, moreTasks, andMore)
+		}
```
mayzhang2000, line 262: If tasks[0] is hook, tasks[1] is pod, the message here is "... and 1 more hook". Variable `andMore` is not accurate.

alexmt: We have two cases where synchronization has to wait for K8S resource: pre-sync/post-sync hooks and sync waves. Pre-sync hooks are executed first, then sync waves ( non-hook resources ), and then post-sync hooks. So we never have a situation while sync is waiting for mixed set of hook and non-hook resources. 

I think it is safe to check if first resource is hook and assume that whole set are hooks.

mayzhang2000: 👍

PR #86, opened 2020-07-10
fix: data race on ctxCompleted

ash2k:
ctxCompleted is read and written from different goroutines, which is a data race.

This change also makes the sleep interruptible.

### DIFF
--- a/pkg/utils/kube/kube.go
+++ b/pkg/utils/kube/kube.go
@@ -18,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
 	kubeyaml "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/apimachinery/pkg/watch"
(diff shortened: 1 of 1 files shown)


pkg/utils/kube/kube.go
```diff
@@ -367,31 +368,20 @@ func GetDeploymentReplicas(u *unstructured.Unstructured) *int64 {
...
-	go func() {
-		select {
-		case <-ctx.Done():
-			ctxCompleted = true
-		case <-stop:
-		}
-	}()
-	for {
+// RetryUntilSucceed keep retrying given action with specified interval until action succeed or specified context is done.
+func RetryUntilSucceed(action func() error, desc string, ctx context.Context, interval time.Duration) {
```
ash2k, line 372: It might be cleaner to have a signature like this:
```go
func RetryUntilSucceed(ctx context.Context, interval time.Duration, desc string, action func() error)
```
context should go first by convention and it's always nicer to pass functions as the last argument because then if it's a closure, it doesn't have more arguments after it, which reads much better.

But I wanted to just fix the bug and not touch the signature here. If you'd like, I can send a follow up PR.

alexmt: Agree, the proposed signature is cleaner and easier to use. This function is used only once in GitOps Engine and twice in Argo CD repo, so it will be easy to migrate to a new signature.

PR #88, opened 2020-07-13
fix: diffing should not fail if known kubernetes resource has invalid fields

alexmt:
The diffing should not fail if known k8s resource has invalid fields.

### DIFF
--- a/pkg/diff/diff_test.go
+++ b/pkg/diff/diff_test.go
@@ -149,6 +149,34 @@ func TestDiff(t *testing.T) {
 	}
 }
 
+func TestDiff_KnownTypeInvalidValue(t *testing.T) {
+	leftDep := newDeployment()
+	leftUn := mustToUnstructured(leftDep)
+	if !assert.NoError(t, unstructured.SetNestedField(leftUn.Object, "badValue", "spec", "revisionHistoryLimit")) {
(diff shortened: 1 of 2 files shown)


pkg/diff/diff_test.go
```diff
@@ -149,6 +149,22 @@ func TestDiff(t *testing.T) {
...
 }
 
+func TestDiff_KnownTypeInvalidValue(t *testing.T) {
+	leftDep := newDeployment()
+	leftUn := mustToUnstructured(leftDep)
+	if !assert.NoError(t, unstructured.SetNestedField(leftUn.Object, "badValue", "spec", "revisionHistoryLimit")) {
+		return
+	}
+
+	diffRes := diff(t, leftUn, leftUn, GetDefaultDiffOptions())
```
mayzhang2000, line 159: This is intentional `leftUn ` is used as input parameter of both `config` and `live`?

alexmt: Yes, this test ensures that diff returns false (not different) for the equal objects even if object has "invalid" fields.

alexmt: implemented an additional use case to ensure that diff correctly handles different resources with invalid fields as well.

PR #90, opened 2020-07-13
feat: create namespace

mayzhang2000:


pkg/sync/sync_context.go
```diff
@@ -781,6 +792,23 @@ func (sc *syncContext) runTasks(tasks syncTasks, dryRun bool) runState {
 					logCtx.Debug("applying")
 					validate := sc.validate && !resourceutil.HasAnnotationOption(t.targetObj, common.AnnotationSyncOptions, common.SyncOptionsDisableValidation)
 					result, message := sc.applyObject(t.targetObj, dryRun, sc.force, validate)
+
+					var nsNotFoundReg = regexp.MustCompile(`namespaces(.*) not found`)
```
alexmt, line 796: I suggest to re-use existing `syncTasks` "framework" to create namespace. The simplest way to do it is to insert `Namespace` pre-sync hook resources in `getSyncTasks`:

https://github.com/argoproj/gitops-engine/blob/344b1bcfc2282b4afc7e8d7475dec00f5d42f542/pkg/sync/sync_context.go#L514-L524


It has several advantages:
* no need to add `CreateNamespace` method in `Kubectl` interface .  
* no need to rely on the error message to detect missing namespace
* the created namespace will be shown in sync status result so user will get more information

mayzhang2000: Great idea!

PR #92, opened 2020-07-14
fix: sync hooks should be deleted after sync phase/wave completion

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/2737

PR changes hook deletion logic so that hooks are deleted after sync phase/wave is completed.

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -10,7 +10,6 @@ import (
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
-	"k8s.io/apimachinery/pkg/api/errors"
 	apierr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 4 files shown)


pkg/sync/sync_context.go
```diff
@@ -341,6 +331,19 @@ func (sc *syncContext) Sync() {
 		return
 	}
 
+	// delete all completed hooks which have appropriate delete policy
+	hooksPendingDeletion := tasks.Filter(func(task *syncTask) bool {
+		return task.isHook() && task.liveObj != nil && !task.running() && task.deleteOnPhaseCompletion()
+	})
+	for _, task := range hooksPendingDeletion {
+		if task.deleteOnPhaseCompletion() {
```
jannfis, line 339: Is this check necessary? It's already part of the task filter above.

pkg/sync/sync_context_test.go
```diff
@@ -593,6 +594,106 @@ func TestRunSyncFailHooksFailed(t *testing.T) {
...
+type resourceNameHealthOverride map[string]health.HealthStatusCode
+
+func (r resourceNameHealthOverride) GetResourceHealth(obj *unstructured.Unstructured) (*health.HealthStatus, error) {
+	if status, ok := r[obj.GetName()]; ok {
+		return &health.HealthStatus{Status: status, Message: "test"}, nil
+	}
+	return nil, nil
+}
+
+func TestRunSync_HooksNotDeletedIfPhaseNotCompleted(t *testing.T) {
```
jannfis, line 606: :+1: :smile:

pkg/sync/sync_context.go
alexmt: nice catch, thank you! the check is not required - removed.

PR #94, opened 2020-07-15
feat: create namespace 

mayzhang2000:


pkg/sync/sync_context.go
```diff
@@ -289,6 +299,19 @@ func (sc *syncContext) Sync() {
 		}
 	}
 
+	//This is to support namespace creation created by pre-sync task.
```
alexmt, line 302: I think it is better to load live namespace resource at the same [place](https://github.com/argoproj/gitops-engine/pull/94/files#diff-f6928024340969ce58a20e045c7dc520R554) where we inject sync task. This way code related to namespace auto-creation won't be spread.

pkg/sync/sync_context.go
```diff
@@ -502,6 +530,35 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
 			task.targetObj = task.targetObj.DeepCopy()
 			task.targetObj.SetNamespace(sc.namespace)
 		}
+		ns := task.targetObj.GetNamespace()
+		if ns != "" {
+			namespaceMap[ns] = ns
+		}
+	}
+
+	if sc.createNamespace && len(namespaceMap) >= 1 {
```
alexmt, line 539: Looks like this will auto-create all namespaces referenced in tasks, right?

Namespace auto-creation should auto-create only target sync namespace (specified in `syncContext.namespace` variable).

pkg/sync/sync_context.go
```diff
@@ -502,6 +530,35 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
...
+		for _, ns := range namespaceMap {
+			existInResources := false
+			//check if namespace exists as part of the resources
+			for _, res := range sc.resources {
+				if res.Target.GetKind() == Namespace && res.Target.GetName() == ns {
```
alexmt, line 544: The `res.Target` might be nil. Add `res.Target !=nil` check please.

alexmt, line 544: We need to check Group too: `res.Target.GetGroup() == ""`

pkg/sync/sync_context.go
```diff
@@ -502,6 +530,35 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
...
+			}
+
+			annotations := make(map[string]string)
+			annotations[common.AnnotationKeyHook] = common.SyncPhasePreSync
+			nsSpec := &v1.Namespace{TypeMeta: metav1.TypeMeta{APIVersion: "v1", Kind: Namespace}, ObjectMeta: metav1.ObjectMeta{Name: ns, Annotations: annotations}}
+			unstructuredObj, err := kube.ToUnstructured(nsSpec)
+			if err == nil {
+				tasks = append(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj})
+			} else {
+				sc.log.Debug(fmt.Sprintf("Failed trying to create a pre-sync task for creating namespace. %s", err))
```
alexmt, line 559: Sync operation should be failed if namespace auto-creation was not successul:
```
sc.setResourceResult(task, task.syncStatus, common.OperationError, fmt.Sprintf("Failed trying to create a pre-sync task for creating namespace. %s", err))
```

pkg/sync/sync_context.go
mayzhang2000: Moved code as close as it can be now. :)

pkg/sync/sync_context.go
```diff
@@ -522,6 +559,24 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
 		}
 	}
 
+	//namespace auto-creation is created by pre-sync task. This task's liveObj is always nil since it did not go through reconciliation.
```
alexmt, line 562: The user might have a custom health check for namespace resource, so the namespace creation task is not necessarily successful when the namespace is created.  I think it is better to just load the actual live resource and at line [536](https://github.com/argoproj/gitops-engine/pull/94/files#diff-f6928024340969ce58a20e045c7dc520R536) an save it in `task.liveObj` field.

pkg/sync/sync_context.go
```diff
@@ -504,6 +514,44 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
...
+			}
+		}
+
+		if isNamespaceCreationNeeded {
+			annotations := make(map[string]string)
+			annotations[common.AnnotationKeyHook] = common.SyncPhasePreSync
+			nsSpec := &v1.Namespace{TypeMeta: metav1.TypeMeta{APIVersion: "v1", Kind: kube.NamespaceKind}, ObjectMeta: metav1.ObjectMeta{Name: sc.namespace, Annotations: annotations}}
+			unstructuredObj, err := kube.ToUnstructured(nsSpec)
+			if err == nil {
+				liveObj, err := sc.kubectl.GetResource(sc.config, unstructuredObj.GroupVersionKind(), unstructuredObj.GetName(), "")
```
alexmt, line 542: Please add `errors.IsNotFound` check. If namespace is not found then we can just set liveObj to nil to indicate that namespace still have to be created.

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/argoproj/pkg v0.0.0-20200102163130-2dd1f3f6b4de
 	github.com/evanphx/json-patch v4.2.0+incompatible
+	github.com/google/gofuzz v1.0.0
```
alexmt, line 8: I think this line is not required either.

PR #96, opened 2020-07-17
fix: kube.SplitYAML should support trailing new lines

alexmt:
PR restores backward compatibility of `kube.SplitYAML` method:

* changes input type to string
* supports YAML that starts with empty space.

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -79,7 +79,7 @@ func (s *settings) parseManifests() ([]*unstructured.Unstructured, string, error
 			if err != nil {
 				return err
 			}
-			items, err := kube.SplitYAML(data)
+			items, err := kube.SplitYAML(string(data))
 			if err != nil {
 				return fmt.Errorf("failed to parse %s: %v", path, err)
 			}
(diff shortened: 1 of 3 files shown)


alexmt:
I've tried to upgrade to most recent gitops-engine version and noticed a problem. Several tests failed because `kube.SplitYAML` failed to parse yaml which starts with new line. E.g.:

```yaml

---
apiVersion: v1
...
``` 

The PR fixes it to keep backward compatibility. At the same time changing input parameter back to `string` to avoid multiple changes in Argo CD.

@ash2k , does this change looks good to you ?

ash2k:
@alexmt Sorry for the breakage! However, I changed the input type to `[]byte` intentionally because this is what the source is all the time, `string` is only used in tests.

I've decided to trace what kubectl does and found that what I've does previously is not the code that it uses. It's confusing that apimachinery has multiple ways to parse YAML. Please have a look at https://github.com/argoproj/gitops-engine/pull/97

alexmt:
No worries! Thank you for PR @ash2k. Closing in favor of https://github.com/argoproj/gitops-engine/pull/97

PR #100, opened 2020-07-21
refactor: remove global semaphore

ash2k:
`rest.Config` has `RateLimiter` that should be used for rate limiting. It's not very good to have global state in a library too.

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -30,35 +30,17 @@ const (
 	clusterSyncTimeout         = 24 * time.Hour
 	watchResourcesRetryTimeout = 1 * time.Second
 	ClusterRetryTimeout        = 10 * time.Second
-)
 
-var (
-	// listSemaphore is used to limit the number of concurrent k8s list queries.
-	// Global limit is required to avoid memory spikes during cache initialization.
+	// Same page size as in k8s.io/client-go/tools/pager/pager.go
+	defaultListPageSize = 500
+	// Prefetch only a single page
+	defaultListPageBufferSize = 1
+	// listSemaphore is used to limit the number of concurrent memory consuming operations on the
+	// k8s list queries results.
+	// Limit is required to avoid memory spikes during cache initialization.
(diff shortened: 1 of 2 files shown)


pkg/cache/cluster.go
```diff
@@ -32,34 +31,6 @@ const (
 	ClusterRetryTimeout        = 10 * time.Second
 )
 
-var (
-	// listSemaphore is used to limit the number of concurrent k8s list queries.
-	// Global limit is required to avoid memory spikes during cache initialization.
-	// The default limit of 50 is chosen based on experiments.
-	listSemaphore = semaphore.NewWeighted(50)
```
alexmt, line 39: Global limit on concurrent lists queries allows avoiding memory spike during cache initialization. Here is corresponding Argo CD ticket: https://github.com/argoproj/argo-cd/issues/3241 

@ash2k , is it possible to use `RateLimiter` to achieve a similar result?

ash2k: @alexmt Hm, no, rate limiter does not allow to achieve what is needed. I've reworked what I'm suggesting to do, please see the commit comment.

alexmt: sorry for delay. looking at PR today/tomorrow

pkg/cache/settings.go
```diff
@@ -75,3 +75,25 @@ func SetConfig(config *rest.Config) UpdateSettingsFunc {
...
+
+// SetListPageBufferSize sets the number of pages to prefetch for list pager.
+func SetListPageBufferSize(listPageBufferSize int32) UpdateSettingsFunc {
+	return func(cache *clusterCache) {
+		cache.listPageBufferSize = listPageBufferSize
+	}
+}
+
+// SetListSemaphore sets the semaphore for list operations.
+// Taking an object rather than a number allows to share a semaphore among multiple caches if necessary.
```
alexmt, line 94: 👍

ash2k:
@alexmt thanks for merging this, I appreciate it! I'll send some more PRs your way soon :)

PR #105, opened 2020-07-23
fix: Namespace auto-creation

mayzhang2000:


james-callahan:
Is it worth adding the option for labels on the created namespace? (to allow tracking which application/team a namespace belongs to?)

mayzhang2000:
Thank you @james-callahan for your review! The intent of this PR is creating a basic namespace. We can create a different issue for the need of having the option for labels for namespace.

mayzhang2000:
> Is it worth adding the option for labels on the created namespace? (to allow tracking which application/team a namespace belongs to?)

Thank you @james-callahan for your review! The intent of this PR is creating a basic namespace. We can create a different issue for the need of having the option for labels for namespace.

pkg/sync/sync_context.go
```diff
@@ -564,9 +564,27 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
 
 	sort.Sort(tasks)
 
+	if sc.createNamespace && sc.namespace != "" {
```
alexmt, line 567: I don't think this is required. I've checked manually and pre-sync state Namespace is executed first already.

Issue #56, opened 2020-06-10
Drop k8s.io/kubernetes dependency


ash2k:
Please consider dropping dependency on `k8s.io/kubernetes` module as it makes the engine really unwieldy to use.

ash2k:
One option would be to copy the bits that are not available in other packages. I'm willing to do the work if maintainers are ok with that. Please let me know.

ash2k:
Hm, I have just discovered that I cannot use the engine because of this dependency. I'm getting the following error when building my project with bazel:
```
ERROR: /private/var/tmp/_bazel_mikhail/6a0c1de97b81a8c5de5b6b9aec394679/external/com_github_argoproj_gitops_engine/pkg/utils/kube/BUILD.bazel:3:11: in go_library rule @com_github_argoproj_gitops_engine//pkg/utils/kube:go_default_library: target '@io_k8s_kubernetes//pkg/kubectl/cmd/auth:go_default_library' is not visible from target '@com_github_argoproj_gitops_engine//pkg/utils/kube:go_default_library'. Check the visibility declaration of the former target if you think the dependency is legitimate
```

This is because https://github.com/kubernetes/kubernetes/blob/07d88617e55b64c243956c7b2032430beb03b159/pkg/kubectl/cmd/auth/BUILD#L16 prohibits to import this package. This is not a problem if Go modules are used, but the package cannot be used via bazel+gazelle (my case). Clearly package authors do not want anyone to depend on it.

ash2k:
I've tried copying the code but ran into these issues:
- `k8s.io/kubernetes/pkg/kubectl/cmd/auth` is a lot of logic, not just trivial helpers.
- looks like creating a custom scheme is not an option because conversion functions are used but they are not exported from k/k.

So I guess I have to look for a workaround...

p.s. I think conversion functions might not be safe to use on the client side. An example: gitops-engine imports kubernetes version `X` and conversion functions for this version. The cluster may be of a different version `Y` and an object of the same group/version/kind might have a new field (I think adding fields within the same version is ok as long as default value matches absent field semantics). This field is unknown to conversion functions from version X and hence will be dropped. Because of that it is not safe to use conversion functions from any other version but `Y`.

ash2k:
https://github.com/kubernetes/enhancements/issues/1020

mdvorak:
Same requirement will be for https://github.com/argoproj/argo-cd, but this needs to be fixed first (as gitops-engine is dependency of argo-cd)

[...]

PR #108, opened 2020-07-28
fix(deps): 🎨  vendor some code from k8s.io/kubernetes to remove depen…

ghostsquad:
…dency

part of #56


ash2k:
I think it's not worth it as this only solves 1% of the problem :) I think our best chance is to talk to k8s maintainers and figure out what the plans are to move this code out of the main repo. I know there is a lot of shuffling going on so maybe these bits are already scheduled to be moved somewhere? If not, this can be a data point in support of such a move.

ghostsquad:
Sounds good.

PR #109, opened 2020-07-29
fix: panic for data type conversion to *unstructured.Unstructured.

mayzhang2000:


ash2k:
I wonder what is the data type that the watch gets when this panic happens? Do you have the message from the panic?

mayzhang2000:
Hi @ash2k, here is the message 
````
Recovered from panic: interface conversion: runtime.Object is *v1.Status, not *unstructured.Unstructured
````

**Comment by ash2k** [comment]
@mayzhang2000 Ok, I see. I think it's not right to just ignore the `Status` object. Getting a `Status` object is most likely an indication that the watch is broken and is not going to send any more events, so ignoring and continuing would lead to a stuck cache, not getting any updates. `Status` should be handled like this https://github.com/kubernetes/client-go/blob/master/tools/watch/retrywatcher.go#L189-L227

Also, I think it'd be best to just use that `RetryWatcher` instead because it not only does what is needed here, it does it better - it supports resuming from the last seen resource version.

WDYT?

File: pkg/cache/cluster.go [hunk 1]
```diff
@@ -479,27 +479,30 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
...
+							}
 						}
-					}
-					if err != nil {
-						c.log.Warnf("Failed to start missing watch: %v", err)
+						if err != nil {
+							c.log.Warnf("Failed to start missing watch: %v", err)
+						}
+					} else {
+						c.log.Warnf("Failed to convert to *unstructured.Unstructured: %v", event.Object)
```
alexmt, line 505: Thank you @ash2k ! The simplest way to retry is to return error here.  e.g. 👍 
```
return fmt.Errorf("Failed to convert to *unstructured.Unstructured: %v", event.Object)
```

mayzhang2000:
Thank you @ash2k and @alexmt for your review!

PR #110, opened 2020-07-31
chore: bump github.com/argoproj/pkg version

ash2k:
Bring in changes from https://github.com/argoproj/pkg/pull/23

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,12 +3,12 @@ module github.com/argoproj/gitops-engine
 go 1.13
 
 require (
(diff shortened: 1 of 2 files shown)


alexmt:
Thank you for PR @ash2k . I agree with your comment in https://github.com/argoproj/pkg/pull/23 about removing `github.com/argoproj/pkg` package dependency.  I think we just did not realize that gitops-engine uses literally just two functions from `argoproj/pkg`. I've created another PR that removes the dependency altogether: https://github.com/argoproj/gitops-engine/pull/111

What do you think? @ash2k , @jannfis ?

ash2k:
Great!

PR #112, opened 2020-08-03
feat: detect PVC StatefulSet ownership

alexmt:
PR introduces dynamic discovery of StatefulSet to PVC ownership. Closes https://github.com/argoproj/argo-cd/issues/2924

### DIFF
--- a/pkg/cache/references.go
+++ b/pkg/cache/references.go
@@ -23,17 +23,19 @@ func (c *clusterCache) resolveResourceReferences(un *unstructured.Unstructured) 
 	var isChildRef func(_ kube.ResourceKey) bool
 	ownerRefs := un.GetOwnerReferences()
 	gvk := un.GroupVersionKind()
+
+	switch {
+
 	// Special case for endpoint. Remove after https://github.com/kubernetes/kubernetes/issues/28483 is fixed
-	if gvk.Group == "" && gvk.Kind == kube.EndpointsKind && len(un.GetOwnerReferences()) == 0 {
+	case gvk.Group == "" && gvk.Kind == kube.EndpointsKind && len(un.GetOwnerReferences()) == 0:
 		ownerRefs = append(ownerRefs, metav1.OwnerReference{
 			Name:       un.GetName(),
 			Kind:       kube.ServiceKind,
 			APIVersion: "v1",
 		})
-	}
 
 	// Special case for Operator Lifecycle Manager ClusterServiceVersion:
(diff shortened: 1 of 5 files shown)


pkg/cache/references.go
```diff
@@ -0,0 +1,130 @@
...
+var (
+	pvcNameRegex = regexp.MustCompile(`.+-(.+)-\d+`)
+)
+
+func (c *clusterCache) resolveResourceReferences(un *unstructured.Unstructured) ([]metav1.OwnerReference, func(kube.ResourceKey) bool) {
+	var isChildRef func(_ kube.ResourceKey) bool
+	ownerRefs := un.GetOwnerReferences()
+	gvk := un.GroupVersionKind()
+	// Special case for endpoint. Remove after https://github.com/kubernetes/kubernetes/issues/28483 is fixed
+	if gvk.Group == "" && gvk.Kind == kube.EndpointsKind && len(un.GetOwnerReferences()) == 0 {
```
ash2k, line 27: nitpick: all these `if`s can probably be turned into a `switch` for better readability.

alexmt: Thank you for the suggestion! Switch is actually easier to read. Updated.

opensourceonly:
@alexmt Do we know if this will be available in which version of Argo CD ?

alexmt:
@opensourceonly  We plan to release is as part of Argo CD v1.7

opensourceonly:
@alexmt i hope 1.7 is not far away to be released

eric-sellers:
Hello @alexmt ,

I would appreciate your confirmation regarding the following.

Since this issue did not make it in to gitops-engine v0.1.3, am I correct in my assumption that Argo CD v1.7 does *not* detect PVC StatefulSet ownership? 

As far as I have experimented by deploying applications that have StatefulSets using ArgoCD v1.6 and v1.7, both show the exact same behaviour, where the PVC appears to be unrelated to the StatefulSet and is marked as "Should be pruned".

If my understanding is correct, the claim that this issue is resolved in v1.7 as announced in the following page is incorrect.

- https://github.com/argoproj/argo-cd/releases
  - feat: detect PVC StatefulSet ownership (#112)

alexmt:
Hello @eric-sellers , it is part of 1.7 but has a bug. Just sent PR with https://github.com/argoproj/gitops-engine/pull/138

eric-sellers:
Hello alexmt,

I see. Thank you for updating me on the latest situation.

opensourceonly:
@alexmt is the bug fix part of a release other than 1.7 ?

PR #102, opened 2020-07-21
feat: use Kubernetes v1.18.6 libraries

ash2k:

PR #113, opened 2020-08-04
fix: specify correct versions

ash2k:
`v1.17.8` is not correct. The right dependency version is `v0.17.8`. I've also bumped a few dependencies.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,18 +6,18 @@ require (
 	github.com/evanphx/json-patch v4.2.0+incompatible
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/pkg/errors v0.9.1
-	github.com/sirupsen/logrus v1.4.2
-	github.com/spf13/cobra v0.0.7
+	github.com/sirupsen/logrus v1.6.0
+	github.com/spf13/cobra v1.0.0
 	github.com/stretchr/testify v1.6.1
-	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
+	golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208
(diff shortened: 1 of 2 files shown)


alexmt:
@ash2k , do we need this PR given that we are migrating to v1.18.6 in https://github.com/argoproj/gitops-engine/pull/102 ?

ash2k:
@alexmt we do not, but I don't know how long it'll take to merge #102, so I opened this one too, which is a smaller change. If you want to merge #102 instead, that's great!

alexmt:
I'm testing https://github.com/argoproj/gitops-engine/pull/102 right now.

Unfortunately, we don't have a lot of e2e tests for gitops-engine. We will work on adding e2e in the future. 

For now, I'm using Argo CD e2e tests (https://github.com/argoproj/argo-cd/pull/4046). I've found a few issues. I plan to send reverse PR to your fork https://github.com/ash2k/gitops-engine/pull/1 :) Still debugging one failed Argo CD e2e test

alexmt:
Closing in favor of https://github.com/argoproj/gitops-engine/pull/102

darshanime:
just for my understanding, since we are `replace`ing all the versions to [the correct `v0.17.8`](https://github.com/argoproj/gitops-engine/blob/f420bb91a1e6345b4f1fbce92e689cb6216c4828/go.mod#L28), is this required?

ash2k:
@darshanime Good question! From https://github.com/golang/go/wiki/Modules#gomod:

> exclude and replace directives only operate on the current (“main”) module. exclude and replace directives in modules other than the main module are ignored when building the main module. 

`gitops-engine` is a library, so when I use it, it's `replace` directives are ignored in my build.

PR #121, opened 2020-08-19
feature: add methods to get annotations

darshanime:
Signed-off-by: darshanime <deathbullet@gmail.com>

required for https://github.com/argoproj/argo-cd/pull/4123

### DIFF
--- a/pkg/utils/kube/kube.go
+++ b/pkg/utils/kube/kube.go
@@ -135,6 +135,24 @@ func GetAppInstanceLabel(un *unstructured.Unstructured, key string) string {
 	return ""
 }
 
+// GetAppInstanceAnnotation returns the application instance name from annotations
+func GetAppInstanceAnnotation(un *unstructured.Unstructured, key string) string {
+	if annotations := un.GetAnnotations(); annotations != nil {
+		return annotations[key]
+	}
+
+	return ""
+}
+
+// GetAppInstanceIdentifier returns the application instance name from annotations else labels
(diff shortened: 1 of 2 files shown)


pkg/utils/kube/kube.go
```diff
@@ -135,6 +135,24 @@ func GetAppInstanceLabel(un *unstructured.Unstructured, key string) string {
 	return ""
 }
 
+// GetAppInstanceAnnotation returns the application instance name from annotations
+func GetAppInstanceAnnotation(un *unstructured.Unstructured, key string) string {
```
ash2k, line 139: This method can be just `un.GetAnnotations()[key]` :) It's safe to index into a `nil` map. https://play.golang.org/p/zvWU5a86Ry_f

pkg/utils/kube/kube.go
```diff
@@ -129,10 +129,21 @@ func MustToUnstructured(obj interface{}) *unstructured.Unstructured {
 
 // GetAppInstanceLabel returns the application instance name from labels
 func GetAppInstanceLabel(un *unstructured.Unstructured, key string) string {
```
alexmt, line 131: @darshanime , looks like `GetAppInstanceLabel` is not used in GitOps engine. Instead of making changes here please move `GetAppInstanceIdentifier` with dependencies to https://github.com/argoproj/argo-cd/pull/4123 and then we can remove  `GetAppInstanceLabel`  from GitOps engine.

alexmt (changes requested):

PR #127, opened 2020-08-24
fix: volumeClaimTemplates are out of sync incorrectly

alexmt:
Fixes https://github.com/argoproj/argo-cd/issues/4126


dpennell:
This still seems to be broken in argocd 1.7.3.

PR #133, opened 2020-08-27
fix: leverage RetryWatcher to watch cluster events and introduce periodical K8S API state resynchronization

alexmt:
We've started seeing cache inconsistency issues recently on both v1.6.0 and v1.7.0 instances. The issue is not intermittent and does not have clear steps to reproduce. I think there are two reasons:

* Argo CD should not rely on 100% watch API correctness. Due to possible racing conditions, some events might be handled in the wrong order or missed. 
* we don't handle all Watch API edge cases and because Argo CD should not rely on 100% watch API correctness. 

PR fixes both issues:

* Introduces periodical ( every 10 mins ) re-synchronization of each monitored API
* Leverages RetryWatcher that simplifies watch API usage and handles API edge cases.

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -20,13 +20,16 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/pager"
+	watchutil "k8s.io/client-go/tools/watch"
 
 	"github.com/argoproj/gitops-engine/pkg/utils/kube"
 )
 
 const (
-	clusterSyncTimeout         = 24 * time.Hour
+	clusterResyncTimeout       = 24 * time.Hour
+	watchResyncTimeout         = 10 * time.Minute
 	watchResourcesRetryTimeout = 1 * time.Second
 	ClusterRetryTimeout        = 10 * time.Second
 
@@ -42,9 +45,8 @@ const (
 )
 
 type apiMeta struct {
-	namespaced      bool
-	resourceVersion string
-	watchCancel     context.CancelFunc
+	namespaced  bool
+	watchCancel context.CancelFunc
 }
 
 // ClusterInfo holds cluster cache stats
@@ -109,7 +111,7 @@ type WeightedSemaphore interface {
 // NewClusterCache creates new instance of cluster cache
 func NewClusterCache(config *rest.Config, opts ...UpdateSettingsFunc) *clusterCache {
 	cache := &clusterCache{
-		resyncTimeout:           clusterSyncTimeout,
+		resyncTimeout:           clusterResyncTimeout,
 		settings:                Settings{ResourceHealthOverride: &noopSettings{}, ResourcesFilter: &noopSettings{}},
 		apisMeta:                make(map[schema.GroupKind]*apiMeta),
 		listPageSize:            defaultListPageSize,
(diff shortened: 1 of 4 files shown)


pkg/cache/cluster.go
```diff
@@ -403,63 +398,77 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
...
 		}
 
-		w, err := resClient.Watch(ctx, metav1.ListOptions{ResourceVersion: info.resourceVersion})
+		w, err := watchutil.NewRetryWatcher(resourceVersion, &cache.ListWatch{
+			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+				return resClient.Watch(ctx, options)
+			},
+		})
+
 		if errors.IsNotFound(err) {
```
ash2k, line 418: If you dig into `watchutil.NewRetryWatcher()` you will see there is no way it can return a NotFound error. In fact, it contacts the remote side asynchronously. I've tried to do the same refactoring as I believe it's very necessary, but this is the blocker I've ran into. I think the case where the watched API went away is definitely a possibility, but not as common. Perhaps there should be some out of bound process that resynchronizes the list of APIs and stops the no longer needed watches (and starts the needed ones, maybe)?

pkg/cache/cluster.go
```diff
@@ -365,22 +362,20 @@ func runSynced(lock sync.Locker, action func() error) error {
...
 				err = fmt.Errorf("Recovered from panic: %+v\n%s", r, debug.Stack())
 			}
 		}()
 
-		err = runSynced(&c.lock, func() error {
-			if info.resourceVersion != "" {
-				return nil
-			}
+		// load API initial state if no resource version provided
+		if resourceVersion == "" {
```
ash2k, line 374: Inner block of this `if` should be in a closure/function to ensure `defer c.listSemaphore.Release(1)` is executed ASAP. Otherwise the semaphore will be held for the duration of the watch which is obviously not what we want. Previously, as you can see, we relied on `runSynced()` closure.

pkg/cache/cluster.go
ash2k: I still think this is an improvement and should be merged :)

pkg/cache/cluster.go
alexmt: Thank you! Nice catch. Fixed

pkg/cache/cluster.go
alexmt: Thank you! We can catch that error in `WatchFunc` implementation. Done.

alexmt:
Thank you for review @ash2k ! PTAL

alexmt:
After removing `resourceVersion` field from `apiMeta` we no longer have racing conditions :) Enabled race detection in unit test by default: https://github.com/argoproj/gitops-engine/pull/133/commits/ede2e4fc7b98acc5b4c3f68a76c0a2670e78be7f

jessesuen:
Interesting note about the RetryWatcher in upstream comments:
```
// RetryWatcher will make sure that in case the underlying watcher is closed (e.g. due to API timeout or etcd timeout)
// it will get restarted from the last point without the consumer even knowing about it.
// RetryWatcher does that by inspecting events and keeping track of resourceVersion.
// Especially useful when using watch.UntilWithoutRetry where premature termination is causing issues and flakes.
// Please note that this is not resilient to etcd cache not having the resource version anymore - you would need to
// use Informers for that.
```

"// Please note that this is not resilient to etcd cache not having the resource version anymore - you would need to
// use Informers for that."

Wonder if our caching problems are somehow related.

pkg/cache/cluster.go
ash2k: I think the fix is not good enough :) For two reasons:
- Previously the semaphore was held during all the lifetime of the `items` variable. Remember, we want to limit memory pressure so it should not only cover the pager but the whole section where `items` variable is used. I think this is important.
- Pager may panic but we recover from panics above. This will lead to semaphore permits leakage. So I think it really should be a closure + `defer` inside. And also I'd get rid of all the panic recover code from everywhere as this cure may be worse than the disease.

alexmt: Thank you @ash2k ! Created follow up PR to address it: https://github.com/argoproj/gitops-engine/pull/135

PR #135, opened 2020-08-28
refactor: ensure list semaphore is released after response is fully processed

alexmt:
Improves changes of https://github.com/argoproj/gitops-engine/pull/133 (see https://github.com/argoproj/gitops-engine/pull/133#discussion_r478791575): ensure that response of every list request is processed completely when limiting semaphore is released.


ash2k (approved):
👍

PR #136, opened 2020-09-01
fix: health status is set to healthy for statefulset with updateStrat…

mayzhang2000:
fixes https://github.com/argoproj/argo-cd/issues/1881

### DIFF
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -332,7 +332,7 @@ func getStatefulSetHealth(obj *unstructured.Unstructured) (*HealthStatus, error)
 	if sts.Spec.UpdateStrategy.Type == appsv1.OnDeleteStatefulSetStrategyType {
 		return &HealthStatus{
 			Status:  HealthStatusHealthy,
-			Message: fmt.Sprintf("updateStrategy is OnDelete type."),
+			Message: fmt.Sprintf("statefulset has %d ready pods", sts.Status.ReadyReplicas),
 		}, nil
 	}
 	if sts.Status.UpdateRevision != sts.Status.CurrentRevision {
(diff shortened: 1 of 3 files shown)


pkg/health/health.go
```diff
@@ -329,6 +329,12 @@ func getStatefulSetHealth(obj *unstructured.Unstructured) (*HealthStatus, error)
 			Message: fmt.Sprintf("partitioned roll out complete: %d new pods have been updated...", sts.Status.UpdatedReplicas),
 		}, nil
 	}
+	if sts.Spec.UpdateStrategy.Type == appsv1.OnDeleteStatefulSetStrategyType {
+		return &HealthStatus{
+			Status:  HealthStatusHealthy,
+			Message: fmt.Sprintf("updateStrategy is OnDelete type."),
```
alexmt, line 335:
```suggestion
			Message: fmt.Sprintf("statefulset has %d ready pods", sts.Status.ReadyReplicas)
```

mayzhang2000:
image

pkg/health/health.go
mayzhang2000: test case is added.

pkg/health/health.go
```diff
@@ -329,6 +329,12 @@ func getStatefulSetHealth(obj *unstructured.Unstructured) (*HealthStatus, error)
 			Message: fmt.Sprintf("partitioned roll out complete: %d new pods have been updated...", sts.Status.UpdatedReplicas),
 		}, nil
 	}
+	if sts.Spec.UpdateStrategy.Type == appsv1.OnDeleteStatefulSetStrategyType {
+		return &HealthStatus{
+			Status:  HealthStatusHealthy,
+			Message: fmt.Sprintf("statefulset has %d ready pods", sts.Status.ReadyReplicas),
```
glasser, line 335: I understand the issue that this change was fixing and it makes sense to me that this counts as "healthy", but what do folks think about iterating on this so that the message lets you know whether or not UpdateRevision matches CurrentRevision in the OnDelete case?

thiagowfx: https://github.com/argoproj/gitops-engine/pull/575

PR #138, opened 2020-09-03
fix: correctly infer ownership references from PVC to StatefulSet

alexmt:
PR improves the fix implemented https://github.com/argoproj/gitops-engine/pull/112 . 

Original changes incorrectly assume that StatefulSet is present in cache when PVCs are discovered. That is incorrect since all resources are loaded in parallel.


pkg/cache/references.go
```diff
@@ -95,9 +74,11 @@ func isStatefulSetChild(un *unstructured.Unstructured) (func(kube.ResourceKey) b
 
 	templates := sts.Spec.VolumeClaimTemplates
 	return func(key kube.ResourceKey) bool {
-		for _, templ := range templates {
-			if strings.HasPrefix(key.Name, fmt.Sprintf("%s-%s-", templ.Name, un.GetName())) {
-				return true
+		if key.Kind == kube.PersistentVolumeClaimKind && key.GroupKind().Group == "" {
```
mayzhang2000, line 77: Should call `mightHaveInferredOwner` function.

pkg/cache/references.go
```diff
@@ -15,12 +14,13 @@ import (
 	"github.com/argoproj/gitops-engine/pkg/utils/kube"
 )
 
-var (
-	pvcNameRegex = regexp.MustCompile(`.+-(.+)-\d+`)
-)
+// mightHaveInferredOwner returns true of given resource might have inferred owners
+func mightHaveInferredOwner(r *Resource) bool {
```
mayzhang2000, line 18: Where is this function called?

mayzhang2000 (approved):
LGTM

PR #139, opened 2020-09-04
Update to Kubernetes v1.19.2

ash2k:
This is not working:
```
--- FAIL: TestDeploymentHealth (0.01s)
    health_test.go:30: 
                Error Trace:    health_test.go:30
                                                        health_test.go:18
                                                        health_test.go:35
                Error:          Received unexpected error:
                                failed to convert *unstructured.Unstructured to *v1.Deployment: converting (v1beta1.Deployment) to (v1.Deployment): unknown conversion
                Test:           TestDeploymentHealth
--- FAIL: TestAPIService (0.01s)
    health_test.go:30: 
                Error Trace:    health_test.go:30
                                                        health_test.go:18
                                                        health_test.go:102
                Error:          Received unexpected error:
                                failed to convert *unstructured.Unstructured to *v1.APIService: converting (v1beta1.APIService) to (v1.APIService): unknown conversion
                Test:           TestAPIService
FAIL
```
The reason is reflection-based conversion has been deprecated and disabled in Kubernetes v1.19.0 (https://github.com/kubernetes/kubernetes/pull/90018). That means health checking code needs to be re-worked to explicitly handle different GVKs for `Deployment` and `APIService`. I think ideally all conversion should be removed in favor of explicit handling. This is more performant, demands less memory and does not depend on the conversion functions in the scheme (which should only be used by Kubernetes itself, not by anything else - that's why they are not in `k8s.io/api`/`k8s.io/client-go`).

Also, it's a poor practice to mutate global variables in another package like that - if some other package does the same there will be a panic. https://github.com/argoproj/gitops-engine/blob/21b78bd366285511267e8f9cf7ec3c3a5fdbd3bc/pkg/health/health.go#L255-L258


ash2k:
I've made the necessary changes to make it work without relying on scheme. Yes, it's ugly and it's code duplication but there is no way around it. However, strictly speaking versions of those resources didn't have to be the same and code could have been different. Eventually old groups/versions will be removed and this code will be deleted so it's temporary. Also, it's good to depend on scheme a bit less.

alexmt:
Thank you @ash2k ! looking

ash2k:
Hm, I think all of the schema-based conversions need to be changed here. The one with failed tests had tests that actually performed vX -> vY conversions. I haven't checked all the other ones, but it's possible they pass only because they "convert" vX -> vX and that works. Also, again, I have checked it only partially, but I think client-go's scheme (the one that is used in the health package), doesn't have any conversion functions registered. Only server-side scheme has conversion functions. So it's likely that they only worked because of the reflection-based code (i.e. poorly), and it has been removed in 1.19.

ash2k:
@alexmt I've updated all health-checking code to use direct unstructured->typed conversion without the scheme. This code should work properly.

pkg/health/health_argo.go
```diff
@@ -0,0 +1,44 @@
+package health
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type nodePhase string
```
ash2k, line 8: I've made the types and constants private since they don't need to be exported from this library

ash2k:
I noticed that there is this set of helpers in kubectl that can probably be used to replace some of the copied code https://github.com/kubernetes/kubectl/blob/master/pkg/polymorphichelpers/rollout_status.go. Let's do it later in a separate PR.

alexmt:
Thanks a lot. Created https://github.com/argoproj/gitops-engine/issues/161 reduce code duplication and try to use https://github.com/kubernetes/kubectl/blob/master/pkg/polymorphichelpers/rollout_status.go

ZiaUrRehman-GBI:
Hey @ash2k, @alexmt can you please let us know in which argoproj version this change will be available?
 Our Argocd version is 1.7.2 and we are facing this issue.

PR #140, opened 2020-09-07
refactor: Return error from context

ash2k:
It's the standard pattern to return `ctx.Err()` when context is done.


alexmt (approved):
LGTM

PR #141, opened 2020-09-17
fix: Support transition from a git managed namespace to auto-create n…

mayzhang2000:
fixes https://github.com/argoproj/argo-cd/issues/4355

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -600,7 +600,7 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
 				} else {
 					sc.log.Infof("Namespace %s is already existed.", sc.namespace)
 					//We need to remove app label to avoid prunning the namespace
-					if liveObj != nil && kube.GetAppInstanceLabel(liveObj, sc.appLabelKey) != "" {
+					if liveObj != nil && sc.appLabelKey != "" && kube.GetAppInstanceLabel(liveObj, sc.appLabelKey) != "" {
 						kube.UnsetLabel(liveObj, sc.appLabelKey)
 						_, err := sc.kubectl.ApplyResource(context.TODO(), sc.rawConfig, liveObj, liveObj.GetNamespace(), cmdutil.DryRunNone, false, true)
 						if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/sync/sync_context.go
```diff
@@ -597,6 +599,14 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
 					tasks = append(tasks, nsTask)
 				} else {
 					sc.log.Infof("Namespace %s is already existed.", sc.namespace)
+					//We need to remove app label to avoid prunning the namespace
+					if liveObj != nil && sc.appLabelKey != "" && kube.GetAppInstanceLabel(liveObj, sc.appLabelKey) != "" {
+						kube.UnsetLabel(liveObj, sc.appLabelKey)
+						_, err := sc.kubectl.ApplyResource(context.TODO(), sc.rawConfig, liveObj, liveObj.GetNamespace(), cmdutil.DryRunNone, false, true)
```
alexmt, line 605: We still need to leverage `syncTasks` to make sure apply is executed only if dry run was successful and permission checks passing. 

Instead of manually executing apply we can create sync task where `target` object is live namespace copy with modified labels.

pkg/sync/sync_context.go
```diff
@@ -252,6 +253,7 @@ type syncContext struct {
 	lock sync.Mutex
 
 	createNamespace bool
+	appLabelKey     string
```
alexmt, line 256: App label is Argo CD specific implementation detail. Instead of hardcoding app label handling here, I suggest providing the function pointer that applies required modification on namespace object

alexmt (changes requested):

pkg/sync/sync_context.go
mayzhang2000: 👍

pkg/sync/sync_context.go
```diff
@@ -597,6 +599,11 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
 					tasks = append(tasks, nsTask)
 				} else {
 					sc.log.Infof("Namespace %s is already existed.", sc.namespace)
+					//We need to remove app label to avoid prunning the namespace
```
alexmt, line 602: This comment seems to be obsolete. Please remote it

alexmt (approved):
LGTM with one minor comment

PR #143, opened 2020-09-23
fix: check resource namespaces are managed

jopit:
Signed-off-by: John Pitman <jpitman@redhat.com>

Related to https://github.com/argoproj/argo-cd/issues/4329

Return an error from `clusterCache.GetManagedLiveObjs` if we find a target object in a namespace that is not managed by the cluster.

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -661,12 +661,33 @@ func (c *clusterCache) IsNamespaced(gk schema.GroupKind) (bool, error) {
 	return false, errors.NewNotFound(schema.GroupResource{Group: gk.Group}, "")
 }
 
+func (c *clusterCache) managesNamespace(namespace string) bool {
+	if len(c.namespaces) == 0 {
+		return true
+	}
+	if len(namespace) == 0 {
+		return true
+	}
+	for _, ns := range c.namespaces {
+		if ns == namespace {
+			return true
+		}
+	}
+	return false
+}
+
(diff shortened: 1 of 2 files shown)


pkg/cache/cluster.go
```diff
@@ -661,13 +661,34 @@ func (c *clusterCache) IsNamespaced(gk schema.GroupKind) (bool, error) {
 	return false, errors.NewNotFound(schema.GroupResource{Group: gk.Group}, "")
 }
 
+func (c *clusterCache) managesNamespace(namespace string) bool {
+	if len(c.namespaces) == 0 {
+		return true
+	}
+	if len(namespace) == 0 {
+		return true
```
alexmt, line 669: I think we should return false here. If `len(namespace) == 0` is true then resource is a cluster-level resource. Cluster level resources are not managed if cluster is connected in namespaced mode so we should return false.

```golang
if namespace == "" {
	return true
}
```

alexmt (changes requested):

pkg/cache/cluster.go
jopit: I've made the suggested change.

jopit:
Hi @alexmt the SonarCloud failure look like it's due to duplicated code in the tests I added.  I think that eliminating the duplication will make the tests harder to understand.  Do you want me to eliminate the duplication anyway?

alexmt (approved):
LGTM

alexmt:
@jopit the SonarCloud check is optional. I think it is ok to have duplication in test

PR #144, opened 2020-09-23
fix: Hook Deletion Policies HookSucceeded should be run after whole H…

mayzhang2000:
…ook succeed and not only Resource succeed

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -345,16 +345,10 @@ func (sc *syncContext) Sync() {
 		return
 	}
 
-	// delete all completed hooks which have appropriate delete policy
+	// collect all completed hooks which have appropriate delete policy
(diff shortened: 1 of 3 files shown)


alexmt (changes requested):
Tried to test it and found a bug. Following pre-sync hook has `HookFailed` delete policy, always fails but never got deleted.

```
apiVersion: batch/v1
kind: Job
metadata:
  name: before
  annotations:
    argocd.argoproj.io/hook: PreSync
    argocd.argoproj.io/hook-delete-policy: HookFailed
spec:
  template:
    spec:
      containers:
      - name: sleep
        image: alpine:latest
        command: ["fail"]
      restartPolicy: Never
  backoffLimit: 0
```

alexmt (approved):
LGTM

mayzhang2000:
fixes https://github.com/argoproj/argo-cd/issues/4384

PR #145, opened 2020-09-25
fix: avoid memory and events spike after forcefully refreshing api cache

alexmt:
The controller cache incorrectly notifies consumer every 10 minutes when re-syncs the K8S API. The PR fixes that issue


mayzhang2000 (approved):
LGTM

PR #154, opened 2020-10-03
fix: Detect unknown fields in invalid specs as OutOfSync 

jgwest:
There are a couple of spots in `diff.go` that are unmarshalling JSON resource data into k8s structs, but in doing so they are silently tossing (invalid) JSON fields that are not present in that struct. In general, this is due to the use of `json.Unmarshal(...)` (whose default implementation silently ignores extra fields) in conjunction with k8s structs that do not contain those extra fields.

The fix is:
- Use `decoder := json.NewDecoder(strings.NewReader(string(data)))` and `decoder.DisallowUnknownFields()` to detect unknown fields when `remarshall`ing
- When `applyPatch`ing against a Kubernetes resource, indirectly apply the result of the scheme default functions to the target JSON, rather than directly marshalling the scheme function results to the k8s resource (which would cause those extra fields to be discarded)

See code comments for details.

Fixes https://github.com/argoproj/argo-cd/issues/1787

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 
 	jsonpatch "github.com/evanphx/json-patch"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


alexmt (approved):
Awesome solution @jgwest ! I did not think it is even possible to support unknown fields. Thank you!

LGTM

PR #156, opened 2020-10-06
Set TLSServerName in NewKubeConfig

ailurarctos:
Update NewKubeConfig to copy the TLSClientConfig.ServerName to the
TLSServerName field of the cluster. This will allow for connecting
to an API server behind a load balancer when the load balancer DNS
name or IP does not match the API server TLS certificate.


alexmt (approved):
LGTM. Thank you!

PR #157, opened 2020-10-09
feat: local crd validation

pepov:
CRDs from Kubernetes version 1.15 support a field `x-kubernetes-preserve-unknown-fields` that affects validation when a CRD introduces a new field and a CR in the same sync defines a value for that field at the same time. 

The above would be a backwards compatible change if validation wouldn't blow up on the missing field in the client. The issue is the same when there are other incompatible changes or even when a new CRD version is introduced, which cannot get applied to the API server because it's CR breaks validation.

We can disable validation for resources already, but it's obviously not ideal.

This PR aims to handle validation against the CRD if it exists in the same sync's targets. A new sync-option annotation `ValidateWithLocalCRD=true` is introduced to make this feature opt-in rather then default.

A test application is prepared for validating the change:
```
argocd app create logging \
    --dest-namespace default --dest-name in-cluster \
    --repo https://github.com/pepov/argocd-integrations.git \
    --path logging --revision 3.5.0
```

Sync the app with `3.5.0` first
```
argocd app sync logging --revision 3.5.0
```

Change revision to `3.6.0-broken` to see the issue without using the annotation:
```
argocd app sync logging --revision 3.6.0-broken
```

Change revision to `3.6.0` to see how setting the annotation should fix the problem:
```
argocd app sync logging --revision 3.6.0
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/argoproj/gitops-engine
 go 1.14
 
 require (
+	github.com/MakeNowJust/heredoc v0.0.0-20170808103936-bb23615498cd
 	github.com/evanphx/json-patch v4.9.0+incompatible
 	github.com/go-logr/logr v0.2.1
 	github.com/golang/mock v1.4.4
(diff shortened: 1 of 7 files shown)


sagikazarmark:
With the latest API version changes in Kubernetes and with the closing deprecation of Helm charts (potentially resulting in upgrades to new versions/operators) this would be really great to have.

pepov:
@alexmt I've changed the code to avoid the extra flag by always validating against a local CRD if it exists in the sync and in case it passed use it as an indicator to retry a failed apply without validation.

PR #159, opened 2020-10-09
feat: Namespace/CRD creation should happen before PreSync phase

alexmt:
PR changes sync tasks sorting so that CRD and Namespace always created before matching CRs or objects in the same namespace.

Closes https://github.com/argoproj/argo-cd/issues/4354


mayzhang2000 (approved):
LGTM

PR #160, opened 2020-10-10
fix: sort endpoint IP addresses before diffing

wtam2018:
**Description**:

This PR fixes https://github.com/argoproj/argo-cd/issues/1816.  For an Endpoint, unsorted/out of lexicographical order IP addresses in the desired state won't match the actual live state and the application won't able to move to "Synced" state.  This PR adds normalize function to sort IP addresses before performing diff so that diff is successful and the application can go to "Synced" state.


alexmt (approved):
LGTM. Thank you @wtam2018 !

pkg/diff/diff.go
```diff
@@ -19,10 +19,17 @@ import (
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/client-go/kubernetes/scheme"
 
+	"k8s.io/kubernetes/pkg/api/endpoints"
```
alexmt, line 22: nitpick: can you please group third party dependencies together?

alexmt:
@wtam2018 resolve  merge conflicts please

ash2k:
It's unfortunate that we're adding more dependencies on the k/k repo. https://github.com/argoproj/gitops-engine/issues/56

PR #162, opened 2020-10-14
refactor: use github.com/go-logr/logr for logging

### LINKED ISSUE #66: gitops-engine should not perform logging on its own
`gitops-engine` API should not perform logging on its own.

We should drop `logrus` dependency from gitops-engine and have consumers specify an event callback instead if they want to process (i.e. log) specific events emitted from the engine. This way, gitops-engine would not interfere with logging requirements from the consumer, i.e. log levels, log format and log output.

If no callbacks are registered, events should be discarded.

ash2k:
Use `github.com/go-logr/logr` instead of Logrus.

Use `k8s.io/klog/v2/klogr` as default implementation because we depend on Kubernetes which uses `klog` anyway. `klog` can be set to use `github.com/go-logr/logr`, which any engine user should probably do.

Closes #66.


pkg/diff/diff_test.go
```diff
@@ -4,49 +4,24 @@ import (
...
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2/klogr"
 	"sigs.k8s.io/yaml"
 )
 
-func captureStdout(callback func()) (string, error) {
```
ash2k, line 26: This is not really needed :D

pkg/diff/diff_test.go
```diff
@@ -56,45 +31,42 @@ func printDiff(result *DiffResult) (string, error) {
...
 	err = ioutil.WriteFile(liveFile, liveData, 0644)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	cmd := exec.Command("diff", liveFile, targetFile)
-	cmd.Stderr = os.Stderr
-	cmd.Stdout = os.Stdout
-	return cmd.Run()
+	return cmd.Output()
```
ash2k, line 69: stdlib already does what we need - removed `captureStdout()`.

pkg/diff/diff_test.go
```diff
@@ -178,10 +150,10 @@ func TestDiff(t *testing.T) {
 	leftDep := newDeployment()
 	leftUn := mustToUnstructured(leftDep)
 
-	diffRes := diff(t, leftUn, leftUn, GetDefaultDiffOptions())
+	diffRes := diff(t, leftUn, leftUn, diffOptionsForTest()...)
 	assert.False(t, diffRes.Modified)
 	ascii, err := printDiff(diffRes)
-	assert.Nil(t, err)
+	require.NoError(t, err)
```
ash2k, line 156: I replaced all `assert.Nil(t, err)` in this file because it was really hard to see what is wrong. Test should stop on first error to avoid executing more assertions and printing not so relevant messages (red herring!).

pkg/utils/tracing/span_test.go
```diff
@@ -1,42 +0,0 @@
-package tracing
```
ash2k, line 1: Hard to test it. I hope it's ok that I removed the test? Otherwise would need to build a testing hardness and I'm somewhat short on time.

pkg/diff/diff.go
```diff
@@ -443,31 +431,33 @@ func normalizeEndpoint(un *unstructured.Unstructured) {
...
 
 	endpoints.SortSubsets(coreEp.Subsets)
 
 	err = v1.Convert_core_Endpoints_To_v1_Endpoints(&coreEp, &ep, nil)
 	if err != nil {
-		log.Warnf("Could not convert from core to vi endpoint type %s: %v", gvk, err)
+		o.log.Error(err, "Could not convert from core to vi endpoint type %s", gvk)
 		return
 	}
-	un.Object, err = runtime.DefaultUnstructuredConverter.ToUnstructured(&ep)
```
ash2k, line 462: I think this is a bug! We don't want to `nil`-out `un.Object` if there was an error, right?

alexmt: I think you are right. 

@wtam2018 , I think implemented fix is correct, right?

pkg/cache/settings.go
```diff
@@ -50,30 +49,21 @@ func SetPopulateResourceInfoHandler(handler OnPopulateResourceInfoHandler) Updat
...
 // SetConfig updates cluster rest config
 func SetConfig(config *rest.Config) UpdateSettingsFunc {
 	return func(cache *clusterCache) {
-		if !reflect.DeepEqual(cache.config, config) {
-			log.WithField("server", cache.config.Host).Infof("Changing cluster config to: %v", config)
-			cache.config = config
-		}
+		cache.config = config
 	}
 }
```
wtam2018, line 68: I guess we don't need these deepequal() checks, correct?

pkg/diff/diff.go
wtam2018: Yes, lgtm.   Thanks for the fix.  @ash2k

pkg/sync/sync_context.go
```diff
@@ -319,7 +326,7 @@ func (sc *syncContext) Sync() {
 			// this must be calculated on the live object
 			healthStatus, err := health.GetResourceHealth(task.liveObj, sc.healthOverride)
 			if err == nil {
-				log.WithFields(log.Fields{"task": task, "healthStatus": healthStatus}).Debug("attempting to update health of running task")
+				sc.log.WithValues("task", task, "healthStatus", healthStatus).V(1).Info("attempting to update health of running task")
```
wtam2018, line 329: Should it be debug or info level logging?

pkg/utils/tracing/span_test.go
wtam2018: @alexmt This test seems to be a keeper to me.   If so, how about creating an issue and document what needs to be done to restore it?

pkg/cache/settings.go
ash2k: No, they were here only for logging, which it seems to me is not very useful.

pkg/utils/tracing/span_test.go
ash2k: @alexmt @wtam2018 I've refactored the tracing "system" and added a test. PTAL.

pkg/utils/kube/ctl.go
```diff
@@ -442,30 +441,9 @@ func (k *KubectlCmd) authReconcile(ctx context.Context, config *rest.Config, kub
 	return strings.Join(out, ". "), nil
 }
 
-func Version() (string, error) {
```
ash2k, line 444: I've removed this function as it's out of line - it's the only thing that calls `kubectl` and is not used in the engine. I hope that's ok. If argo needs it, it can be copied. I hope you agree with the idea of keeping this library clean and small.

pkg/utils/tracing/span.go
```diff
@@ -1,51 +0,0 @@
...
-
-type Span struct {
-	operationName string
-	baggage       map[string]interface{}
-	start         time.Time
-}
-
-func (s Span) Finish() {
-	if enabled {
-		logger.WithValues(baggageToVals(s.baggage)).
```
ash2k: A bug! `WithValues()` accepts a list of key-value pairs but here code passes a slice of interfaces as the key and does not pass any value. Did it even work? The right thing to do here is to expand the slice using `...`, which is what I did above in the other file. `s.logger.WithValues(baggageToVals(s.baggage)...)`

pkg/utils/tracing/span_test.go
wtam2018: It looks pretty neat.   

Tracking is turned off by default and it is enabled by setting the env var below.   Should we keep this behavor?

`enabled = os.Getenv("ARGOCD_TRACING_ENABLED") == "1"`

ash2k: Gitops engine is a library so I think ideally it should not depend on any environment variables. It should be up to the integrating application to configure it how it wants to. Argo can easily check the environment variable and inject the logging tracer into the library if tracing is enabled.

pkg/cache/cluster.go
```diff
@@ -652,7 +657,7 @@ func (c *clusterCache) IterateHierarchy(key kube.ResourceKey, action func(resour
 					return strings.Compare(key1.String(), key2.String()) < 0
 				})
 				child := children[0]
-				action(child, nsNodes)
+				action(nil, child, nsNodes)
 				child.iterateChildren(nsNodes, map[kube.ResourceKey]bool{res.ResourceKey(): true}, action)
```
alexmt, line 661: I think it is better to keep `circular dependency` error internal to gitops-engine. Can you remove `error` parameter `cluster.IterateHierarchy` method and keep it in `resource.iterateChildren` only. E.g. : 

```
child.iterateChildren(nsNodes, map[kube.ResourceKey]bool{res.ResourceKey(): true}, func(err error, child *Resource, namespaceResources map[kube.ResourceKey]*Resource) {
	if err != nil {
		c.log.V(2).Info(err.Error()) 
	}
	action(child, namespaceResources)
})
```

alexmt (changes requested):
Thank you @ash2k ! 

Added one minor comment.

pkg/utils/errors/errors.go
```diff
@@ -1,9 +1,10 @@
 package errors
 
 import (
```
alexmt, line 3: Package `pkg/utils/errors` is used only in `agent/main.go` which only demonstrates how to use gitops-engine but it used 400+ times in Argo CD. 

I think it indicates that `pkg/utils/errors` don't belong to `gitops-engine` at all .

Can you please remove `pkg/utils/errors` completely and stop using it  `agent/main.go`?

alexmt (changes requested):

pkg/utils/io/io.go
```diff
@@ -46,8 +46,8 @@ func NewCloser(close func() error) Closer {
 
 // Close is a convenience function to close a object that has a Close() method, ignoring any errors
 // Used to satisfy errcheck lint
-func Close(c Closer) {
+func Close(c Closer, log logr.Logger) {
```
alexmt, line 49: The `io.Close` is used extremely frequently in Argo CD. It would be really painful to add one more parameter. Instead can we global package level logger instance ?

alexmt (changes requested):
Found couple more changes (add logger parameter to `errors` package and `io.Close` function ) that make it really fo Argo CD

In order to find problematic breaking changes I had to try to compile and test Argo CD with new version [here](https://github.com/argoproj/argo-cd/pull/4652)

So I've implemented proposed solution and sent PR to your branch:

https://github.com/ash2k/gitops-engine/pull/3

pkg/utils/io/io.go
ash2k: Thanks for the review! I've made the changes you have suggested, except for this one. May I suggest to copy this tiny function into Argo-CD instead of introducing a global logger just for this case? Would be really nice to avoid any globals in the library.

alexmt: Sounds good :) @ash2k can you please remove `Close` function as well as `Closer` interface with `inlineCloser` ? I'm working on moving them to Argo CD

ash2k: @alexmt Done! Looks better now in my opinion! Thanks.

alexmt (approved):
Thank you! LGTM

PR #168, opened 2020-10-21
fix: Data race between gitops-engine's pkg/cache/cluster.go and itself, on Argo CD startup (#4627)

jgwest:
The fix is straightforward, as far as these things go: 
- Introduce a new lock, `clusterCacheSync.lock` which is only used for accessing the `syncStatus` fields of the cluster cache.
- Move fields into a `clusterCacheSync` struct which is managed by that lock
- This new lock is much cheaper to acquire than the cluster lock, since it does not block on the same heavyweight operations which use the cluster lock (with the one pre-existing exception of `EnsureSynced`'s resync operation)
- Add lock acquisition to code that uses `syncStatus`
- Add comment describing rules for acquiring `syncStatus` so that we don't deadlock.

**Parent issue**:
https://github.com/argoproj/argo-cd/issues/4627

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -114,7 +114,6 @@ type WeightedSemaphore interface {
 func NewClusterCache(config *rest.Config, opts ...UpdateSettingsFunc) *clusterCache {
 	log := klogr.New()
 	cache := &clusterCache{
-		resyncTimeout:      clusterResyncTimeout,
 		settings:           Settings{ResourceHealthOverride: &noopSettings{}, ResourcesFilter: &noopSettings{}},
 		apisMeta:           make(map[schema.GroupKind]*apiMeta),
 		listPageSize:       defaultListPageSize,
@@ -127,7 +126,10 @@ func NewClusterCache(config *rest.Config, opts ...UpdateSettingsFunc) *clusterCa
 			Log:    log,
 			Tracer: tracing.NopTracer{},
 		},
-		syncTime:                nil,
+		syncStatus: clusterCacheSync{
+			resyncTimeout: clusterResyncTimeout,
+			syncTime:      nil,
+		},
 		resourceUpdatedHandlers: map[uint64]OnResourceUpdatedHandler{},
 		eventHandlers:           map[uint64]OnEventHandler{},
 		log:                     log,
@@ -139,9 +141,8 @@ func NewClusterCache(config *rest.Config, opts ...UpdateSettingsFunc) *clusterCa
 }
 
 type clusterCache struct {
-	resyncTimeout time.Duration
-	syncTime      *time.Time
-	syncError     error
+	syncStatus clusterCacheSync
+
 	apisMeta      map[schema.GroupKind]*apiMeta
 	serverVersion string
 	apiGroups     []metav1.APIGroup
@@ -170,6 +171,17 @@ type clusterCache struct {
 	populateResourceInfoHandler OnPopulateResourceInfoHandler
 	resourceUpdatedHandlers     map[uint64]OnResourceUpdatedHandler
 	eventHandlers               map[uint64]OnEventHandler
+}
+
+type clusterCacheSync struct {
+	// When using this struct:
+	// 1) 'lock' mutex should be acquired when reading/writing from fields of this struct.
+	// 2) The parent 'clusterCache.lock' does NOT need to be owned to r/w from fields of this struct (if it is owned, that is fine, but see below)
+	// 3) To prevent deadlocks, do not acquire parent 'clusterCache.lock' after acquiring this lock; if you need both locks, always acquire the parent lock first
+	lock          sync.Mutex
+	syncTime      *time.Time
+	syncError     error
+	resyncTimeout time.Duration
 }
 
 // OnResourceUpdated register event handler that is executed every time when resource get's updated in the cache
@@ -312,7 +324,11 @@ func (c *clusterCache) setNode(n *Resource) {
 func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
-	c.syncTime = nil
+
+	c.syncStatus.lock.Lock()
+	c.syncStatus.syncTime = nil
+	c.syncStatus.lock.Unlock()
+
 	for i := range c.apisMeta {
 		c.apisMeta[i].watchCancel()
 	}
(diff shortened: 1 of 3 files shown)


pkg/cache/cluster.go
```diff
@@ -134,8 +136,9 @@ func NewClusterCache(config *rest.Config, opts ...UpdateSettingsFunc) *clusterCa
 
 type clusterCache struct {
 	resyncTimeout time.Duration
-	syncTime      *time.Time
-	syncError     error
+
+	syncStatus *clusterCacheSync
```
ash2k, line 140: It does not need to be a pointer, does it?

pkg/cache/cluster.go
```diff
@@ -319,11 +336,13 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
 }
 
 func (c *clusterCache) synced() bool {
```
ash2k, line 338: Would it make sense to make this a method on `clusterCacheSync` and move locking here?

pkg/cache/cluster.go
```diff
@@ -596,23 +615,32 @@ func (c *clusterCache) sync() error {
...
+		syncError := syncStatus.syncError
+		syncStatus.lock.Unlock()
+		return syncError
 	}
+	syncStatus.lock.Unlock() // release the lock, so that we can acquire the parent lock (see struct comment re: lock acquisition ordering)
 
 	c.lock.Lock()
+	syncStatus.lock.Lock()
 	defer c.lock.Unlock()
+	defer syncStatus.lock.Unlock()
```
ash2k, line 632:
```suggestion
	c.lock.Lock()
	defer c.lock.Unlock()
	syncStatus.lock.Lock()
	defer syncStatus.lock.Unlock()
```
Reads better and is more logical? Same in another method.

pkg/cache/cluster.go
```diff
@@ -596,23 +615,32 @@ func (c *clusterCache) sync() error {
...
+
 	// before doing any work, check once again now that we have the lock, to see if it got
 	// synced between the first check and now
 	if c.synced() {
-		return c.syncError
+		return syncStatus.syncError
 	}
 	err := c.sync()
```
ash2k, line 639: So we are running this expensive method while holding the new lock here. What is the point of the new lock then if it's held here? Just trying to understand.

ash2k (commented):
Hey, thanks a lot for the PR! I've been meaning to fix this data race but never got around to it. I've left some comments below.

pkg/cache/cluster.go
jgwest: A good question.... my main goal here was to preserve the behaviour of the existing code, while making it data race safe.

The situation the original code was (AFAICT) trying to avoid is forcing external calls to `EnsureSynced()` (which ensures that the cluster cache is synced to within the given timeframe) to block on expensive operations (eg cluster I/O) that might be in progress at the same time.

The `clusterCache.lock` lock is generally reserved for code that will (necessarily) take a while, due to the need to wait for cluster I/O:
- refreshing the cluster cache
- starting/cancelling watches
- calling resourceUpdate/event handlers
- calling other k8s cluster APIs

In contrast, `clusterCacheSync.lock` is generally reserved for code that desires fast access:
- The first half ("happy path") of `EnsureSynced()`
    - comparing `syncTime` with `resyncTimeout` (`synced()`)
- setting and retrieving the field values 

The _only_ case where `clusterCacheSync.lock` is held on a slow operation:
- the second half of `EnsureSynced()` (the cache invalid, non "happy path" part of `EnsureSynced()`)
    -  And this is consistent with the expected behaviour of `EnsureSynced()` by callers... if the cluster is not synced, we do not want to return from this method until it _is_ synced.

So ideally, in the vast majority of cases, calls to `EnsureSynced()` by consuming code will be able to complete by only acquiring the fast lock (`clusterCacheSync.lock`), and thus the calling code will not need to block on the slower lock.

In the minority of cases (when the cache has not yet synced or the cache sync has expired) the full lock will need to be acquired.

pkg/cache/cluster.go
jgwest: Done... I wanted to avoid accidental [copying of the mutex](https://eli.thegreenplace.net/2018/beware-of-copying-mutexes-in-go/), but it's true this way is cleaner.

pkg/cache/cluster.go
jgwest: Done... I wanted to avoid accidental [copying of the mutex](https://eli.thegreenplace.net/2018/beware-of-copying-mutexes-in-go/), but it's true this way is cleaner.

pkg/cache/cluster.go
jgwest: Done, definitely an improvement.

jgwest: Done... I have moved `synced()` into a method on `clusterCacheSync`, but there are still cases where the clusterCacheSync lock needs be held outside this method.

pkg/cache/cluster.go
jgwest: Done (in both places)... my mild personal preference was to have it the other way, but they are semantically equivalent, and so I'm happy if you're happy 😄

pkg/cache/cluster.go
ash2k: I see. I wonder if we can just use read lock of `lock` to protect reads. That would simplify things quite a bit, but I'm not clear whether it would or would not create any problems. It seems to me that logically `synched` means all the state is up to date. So I'm not sure how this check can be performed out of bound, not holding the main lock. To put it differently, how can we know if it's synchronized if a mutative operation is happening right now? We need to wait for it to finish, don't we?

@alexmt Could you clarify how this should work please?

jgwest: I would argue that the alternative, using a read lock, is unfortunately more complex. 

Why? With an RWLock, write access on that lock will block all the reads, so in order to judge the cost  of using a read lock to fix this data race, we first need to look at who acquires the write access.

So the main acquirer here is `runSynced()` (in the same file), which is used to run a bunch of different functions... Unfortunately, the contained functions contain indirect calls to code that is not only outside this file, but outside of the package, and even outside of gitops-engine entirely. So the scope and potential cost of write lock ownership is broad.

Doing some code tracing: in a nearly all of the cases where the write lock is acquired by `runSynced()`, `onNodeRemoved` and `onNodeUpdated` are called. These two functions will then call the `resourceUpdateHandlers` for the `clusterCache`.

The `resourceUpdateHandlers` are functions which are injected by `OnResourceUpdated`, and `OnResourceUpdated` is called by consuming code that is both inside and outside gitops engine. 
- In gitops engine: It appears that a call to `syncContext.Sync()` will block in a resourceUpdateHandler, so this alone makes the write lock very heavy.
- Outside gitops engine: `liveStateCache.onObjectUpdated` may involve cluster I/O, and we would probably want to avoid blocking a lock on network I/O where possible.


In any case, in order to judge how heavy this lock is, you have to do a LOT of code tracing across packages/components, and (extra unfortunately) you would need to make sure that the desired properties (not holding the write lock for too long) of these packages/components never change in a way that would make the write lock (in this PR) more expensive.

So IMHO I don't think it's feasible to rely on calling code not to break us in this case, and that unfortunately means the introduction of the second lock, as done in the PR.

Though, in contrast to the above described complexity:
- This PR sets a strict set of locking rules that only apply to a single field in a struct, and only apply within a single file.
- This PR preserves the existing concurrency properties of the code (eg the code should work exactly the same as before, neither deadlocking or blocking in unexpected ways)

ash2k: @jgwest Thanks for a very detailed response! 👍 I think it makes sense, thanks for investigating the issue so thoroughly.

alexmt: Ops, missed this PR. Thank you @jgwest ! Looking

jgwest:
I've updated commits to fix DCO check, and I've confirmed the original issue still exists.

jgwest:
Hey @ash2k @alexmt , y'all still interested in merging this? Any other questions/comments?

ash2k (approved):

ash2k:
@jgwest Hey, thanks for the ping. I've approved but I don't have merge permissions.

alexmt (approved):
Apologies for the delay. LGTM. Just in case running Argo CD e2e tests and then merging: https://github.com/argoproj/argo-cd/actions/runs/481146899

PR #175, opened 2020-10-27
refactor: allow PatchResource to accept subresource

jessesuen:
I will need to start using PatchResource with the `status` subresources in order to fix https://github.com/argoproj/argo-cd/issues/4670. This change updates method signature to carry forward the subresources string array.


alexmt (approved):
LGTM

PR #177, opened 2020-10-30
feat: introduce SyncWaveHook callbacks which are invoked after applying each sync wave

jessesuen:
This is groundwork for fixing https://github.com/argoproj/argo-cd/issues/4669. It introduces a SyncWaveHook callback function, which is invoked after successfully applying each sync wave. Argo CD will leverage this by sleeping for a few seconds between each wave.

Also this fixes some logging issues after switching to the new logr.

### DIFF
--- a/pkg/sync/common/types.go
+++ b/pkg/sync/common/types.go
@@ -29,6 +29,11 @@ type PermissionValidator func(un *unstructured.Unstructured, res *metav1.APIReso
 
 type SyncPhase string
 
+// SyncWaveHook is a callback function which will be invoked after each sync wave is successfully
+// applied during a sync operation. The callback indicates which phase and wave it had just
+// executed, and whether or not that wave was the final one.
+type SyncWaveHook func(phase SyncPhase, wave int, final bool) error
+
 const (
 	SyncPhasePreSync  = "PreSync"
 	SyncPhaseSync     = "Sync"
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -414,8 +414,10 @@ func (sc *syncContext) Sync() {
 	if sc.syncWaveHook != nil && runState != failed {
 		err := sc.syncWaveHook(phase, wave, finalWave)
 		if err != nil {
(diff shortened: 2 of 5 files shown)


pkg/sync/common/types.go
```diff
@@ -29,6 +29,11 @@ type PermissionValidator func(un *unstructured.Unstructured, res *metav1.APIReso
 
 type SyncPhase string
 
+// SyncWaveHook is a callback function which will be invoked after each sync wave is successfully
+// applied during a sync operation. The callback indicates which phase and wave it had just
+// executed, and whether or not that wave was the final one.
+type SyncWaveHook func(phase SyncPhase, wave int, final bool) error
```
jessesuen, line 35: Other name I considered was: SyncWaveCallback

pkg/sync/sync_context.go
```diff
@@ -400,6 +410,15 @@ func (sc *syncContext) Sync() {
 
 	sc.log.WithValues("tasks", tasks).V(1).Info("Wet-run")
 	runState := sc.runTasks(tasks, false)
+
+	if sc.syncWaveHook != nil && runState != failed {
```
mayzhang2000, line 414: Here if `runState` is failed, `SyncWaveHook` is not called. I wonder if there is a use case that `SyncWaveHook `needs the flexibility to do different things based on the `runState`? Do we want the input parameter of SyncWaveHook also includes runState?

mayzhang2000 (approved):
LGTM

alexmt (approved):
LGTM

pkg/sync/sync_context.go
```diff
@@ -400,6 +410,15 @@ func (sc *syncContext) Sync() {
...
+		err := sc.syncWaveHook(phase, wave, finalWave)
+		if err != nil {
+			sc.setOperationPhase(common.OperationFailed, fmt.Sprintf("SyncWaveHook failed: %v", err))
+			runState = failed
```
jessesuen, line 418: I just noticed the operation message will get clobbered by the one below. Will fix it before merging.

PR #183, opened 2020-11-13
fix: WithInitialState should require start time to support generateName hooks properly

jessesuen:
This is a follow up to https://github.com/argoproj/gitops-engine/pull/180.

When resuming a sync operation, we need startTime to be stable from previous reconciliation, because startTime is used in sync hooks which use generateName(). Without this fix, we get into a continuous loop where we create infinite number of hooks when they use generateName.


mayzhang2000 (approved):
LGTM

PR #190, opened 2020-12-09
fix: HPA health check is making incorrect assumption on annotations

mayzhang2000:
fixes https://github.com/argoproj/argo-cd/issues/4995

This change is based on https://github.com/argoproj/argo-cd/pull/4915#issuecomment-740246855. 
1. It moved HPA health check from lua script to gitops-engine
2. If annotation `autoscaling.alpha.kubernetes.io/conditions` does not exist, status is healthy


pkg/health/health_hpa.go
```diff
@@ -0,0 +1,69 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type hpaConditions struct {
+	Type    string
+	Reason  string
+	Message string
+}
+
+func is_degraded(condition *hpaConditions) bool {
```
alexmt, line 14: I think we don't use `_`. Rename to `isDegraded`?

pkg/health/health_hpa.go
```diff
@@ -0,0 +1,69 @@
...
+	return false
+}
+
+func getAutoscalingv1HPAHealth(obj *unstructured.Unstructured) (*HealthStatus, error) {
+	healthyStatus := &HealthStatus{
+		Status:  HealthStatusHealthy,
+		Message: "",
+	}
+
+	annotation, ok := obj.GetAnnotations()["autoscaling.alpha.kubernetes.io/conditions"]
```
alexmt, line 35: As https://github.com/argoproj/argo-cd/issues/4995 mentions, HPA might have conditions stored under `status`. We need to check both annotation and conditions field in the status.

alexmt (approved):
LGTM

PR #195, opened 2020-12-21
refactor: Kubernetes v1.20.1

ash2k:


pkg/cache/cluster_test.go
```diff
@@ -44,49 +48,6 @@ func strToUnstructured(jsonStr string) *unstructured.Unstructured {
 
 var (
 	testCreationTime, _ = time.Parse(time.RFC3339, "2018-09-20T06:47:27Z")
-	testPod             = strToUnstructured(fmt.Sprintf(`
```
ash2k, line 47: Because of https://github.com/kubernetes/kubernetes/pull/96020 fake dynamic client now wants either typed objects+proper schema or no schema and explicit list types mapping. The later is discouraged, so I refactored the test to use typed objects.

pkg/cache/cluster_test.go
```diff
@@ -107,13 +68,8 @@ var (
...
 
-func newCluster(objs ...*unstructured.Unstructured) *clusterCache {
-	runtimeObjs := make([]runtime.Object, len(objs))
-	for i := range objs {
-		runtimeObjs[i] = objs[i]
-	}
-	scheme := runtime.NewScheme()
-	client := fake.NewSimpleDynamicClient(scheme, runtimeObjs...)
+func newCluster(t *testing.T, objs ...runtime.Object) *clusterCache {
+	client := fake.NewSimpleDynamicClient(scheme.Scheme, objs...)
```
ash2k, line 72: Proper scheme here now.

pkg/cache/cluster_test.go
```diff
@@ -142,8 +98,12 @@ func newCluster(objs ...*unstructured.Unstructured) *clusterCache {
 		Meta:                 metav1.APIResource{Namespaced: true},
 	}}
 
-	return NewClusterCache(
+	cache := NewClusterCache(
 		&rest.Config{Host: "https://test"}, SetKubectl(&kubetest.MockKubectlCmd{APIResources: apiResources, DynamicClient: client}))
+	t.Cleanup(func() {
+		cache.Invalidate()
```
ash2k, line 104: Stop goroutines we started.

pkg/cache/cluster_test.go
```diff
@@ -155,20 +115,33 @@ func getChildren(cluster *clusterCache, un *unstructured.Unstructured) []*Resour
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "helm-guestbook2",
+			Namespace: "default2",
+		},
+	}
+
+	cluster := newCluster(t, obj1, obj2)
 	err := cluster.EnsureSynced()
-	assert.Nil(t, err)
+	require.NoError(t, err)
```
ash2k, line 141: It was annoying to debug because of a lot of noise. This ensures we see only the first failure and the test stops here, not generating any noise.

pkg/cache/cluster_test.go
```diff
@@ -155,20 +115,33 @@ func getChildren(cluster *clusterCache, un *unstructured.Unstructured) []*Resour
...
+
+	cluster.lock.Lock()
+	defer cluster.lock.Unlock()
```
ash2k, line 144: Tests access internal cache fields while the goroutines are still running and this leads to data races. Added locking in a bunch of tests.

pkg/diff/diff.go
```diff
@@ -518,20 +516,9 @@ func normalizeEndpoint(un *unstructured.Unstructured, o options) {
...
 	}
-	var coreEp core.Endpoints
-	err = v1.Convert_v1_Endpoints_To_core_Endpoints(&ep, &coreEp, nil)
-	if err != nil {
-		o.log.Error(err, "Could not convert from v1 to core endpoint type %s", gvk)
-		return
-	}
 
-	endpoints.SortSubsets(coreEp.Subsets)
+	endpoints.SortSubsets(ep.Subsets)
```
ash2k, line 520: The method has been moved, no need to convert back and forth now 🥳

pkg/utils/kube/ctl.go
```diff
@@ -23,9 +23,9 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/kubectl/pkg/cmd/apply"
+	"k8s.io/kubectl/pkg/cmd/auth"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/scheme"
-	"k8s.io/kubernetes/pkg/kubectl/cmd/auth"
```
ash2k, line 28: Package has been moved, one less dependency on `k8s.io/kubernetes` 🎉

pkg/utils/kube/scheme/scheme.go
```diff
@@ -23,7 +23,6 @@ import (
 	_ "k8s.io/kubernetes/pkg/apis/policy/install"
 	_ "k8s.io/kubernetes/pkg/apis/rbac/install"
 	_ "k8s.io/kubernetes/pkg/apis/scheduling/install"
-	_ "k8s.io/kubernetes/pkg/apis/settings/install"
```
ash2k, line 26: `PodPreset` is no more.

pkg/utils/kube/ctl.go
alexmt: that is awesome!

alexmt (approved):
Changes looks good, also run Argo CD e2e.

Thank you! LGTM

PR #203, opened 2021-01-06
feat: prune last

mayzhang2000:
fixes: #5080.
If sync option has `PruneLast=true`, or individual resource has annotation `argocd.argoproj.io/sync-options: PruneLast=true`, these prune tasks get assigned with a new sync wave.

the new sync wave is the last sync wave of non-prune tasks which is in sync phase + 1, so that these resources will be pruned after all sync phase resources are synced and healthy.


alexmt (approved):
LGTM.

PR #213, opened 2021-01-28
feat: selective sync

kshamajain99:

issue: https://github.com/argoproj/argo-cd/issues/3877
argo-cd PR: https://github.com/argoproj/argo-cd/pull/5347

This feature applies only out-of-sync resources when `ApplyOutOfSync=true` option is provided by user. 

#### Test performed:

Scenario: For an application with 10 deployments and 1 svc, deleting svc and syncing entire application (performing same action)

With `ApplyOutOfSync=true` enabled
```
INFO[0040] sync/terminate complete              application=guestbook duration=507.61639ms syncId=00001-NPWVT
```

Without `ApplyOutOfSync=true` enabled
```
INFO[0114] sync/terminate complete              application=guestbook duration=1.983608856s syncId=00002-orWNj
```

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -2,6 +2,7 @@ package sync
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"sort"
 	"strconv"
@@ -23,6 +24,7 @@ import (
 	"k8s.io/klog/v2/klogr"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 
+	"github.com/argoproj/gitops-engine/pkg/diff"
 	"github.com/argoproj/gitops-engine/pkg/health"
 	"github.com/argoproj/gitops-engine/pkg/sync/common"
 	"github.com/argoproj/gitops-engine/pkg/sync/hook"
@@ -129,6 +131,18 @@ func WithPruneLast(enabled bool) SyncOpt {
 	}
 }
 
+// WithResourceModificationChecker sets resource modification result
+func WithResourceModificationChecker(enabled bool, diffResults *diff.DiffResultList) SyncOpt {
+	return func(ctx *syncContext) {
+		ctx.applyOutOfSyncOnly = enabled
+		if enabled {
+			ctx.modificationResult = groupDiffResults(diffResults)
+		} else {
+			ctx.modificationResult = nil
+		}
+	}
+}
+
 // WithNamespaceCreation will create non-exist namespace
 func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool) SyncOpt {
 	return func(ctx *syncContext) {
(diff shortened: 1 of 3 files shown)


pkg/sync/sync_context.go
```diff
@@ -129,6 +129,14 @@ func WithPruneLast(enabled bool) SyncOpt {
 	}
 }
 
+// WithResourceModificationChecker sets resource modification checker
+func WithResourceModificationChecker(enabled bool, checker common.ModificationChecker) SyncOpt {
```
alexmt, line 133: Instead of asking consumer to re-comparing resources I would suggest passing `*diff.DiffResultList` : 

```
func WithApplyOutOfSyncOnly(enabled bool, diffResult *diff.DiffResultList) {
	if enabled {
		ctx.applyOutOfSyncDiffs = nil
	} else {
		ctx.applyOutOfSyncDiffs = diffResult
	}
}
```
The `*diff.DiffResultList` is always available prior to syncing.

alexmt (changes requested):

alexmt (approved):
LGTM! Please sign commit so we can merge it

peterbosalliandercom:
@alexmt Does this option influence the post-sync hooks in some way?

alexmt:
Hello @peterbosalliandercom . No, it should only skip normal resources sync if no resource is already in sync? Did you notice an unexpected difference?

peterbosalliandercom:
> Hello @peterbosalliandercom . No, it should only skip normal resources sync if no resource is already in sync? Did you notice an unexpected difference?

In the past when doing a sync for only out-of-sync items the post sync hook did not trigger thats why. But in this case it should also perform a post sync hook right?

peterbosalliandercom:
@alexmt Is there a reason why this selective sync is not enabled by default? I cannot think of a reason why you want a full sync everytime.

peterbosalliandercom:
Is there a way to see when a selective sync has been done?

kshamajain99:
@peterbosalliandercom We didn't want to change the existing behaviour of how argoCD is performing sync operation that is why this option is not enabled by default. 

>Is there a way to see when a selective sync has been done?

You can check argoCD controller log to see what all tasks are being synced and whether all or partial tasks are being synced.

peterbosalliandercom:
> @peterbosalliandercom We didn't want to change the existing behaviour of how argoCD is performing sync operation that is why this option is not enabled by default.
> 
> > Is there a way to see when a selective sync has been done?
> 

Is there a way to enable by default without amending all application resources?

kshamajain99:
@peterbosalliandercom You can add this sync option to your application manifest
https://argoproj.github.io/argo-cd/user-guide/sync-options/#selective-sync

peterbosalliandercom:
> @peterbosalliandercom You can add this sync option to your application manifest
> https://argoproj.github.io/argo-cd/user-guide/sync-options/#selective-sync

Yes I understand but then I have to update all application manifests. Is there a way to set it default on cluster level?

PR #219, opened 2021-02-17
fix: when there are namespace resources, sync them first.

mayzhang2000:

When resources contain namespace resource, along with resource `ClusterRoleBinding` which is a PreSync hook, resource `ClusterRoleBinding` is synced prior to namespace resource. This is due to `ClusterRoleBinding` resource does not have namespace field specified. 

This PR is to fix when there are namespace resources, we should sync namespace resources first.


alexmt (changes requested):
@mayzhang2000 , are you referring to `ingress-nginx` ClusterRole in https://raw.githubusercontent.com/jannfis/argocd-example-apps/master/ingress/namespace/deploy.yaml ? 


This resource does not have hard-coded namespace and don't have to be synced before Namespace. The goal of https://github.com/argoproj/argo-cd/issues/5514 was to make sure that namespace resource is synced before resources that **reside in that namespace**, not just create namespace first all the time.

PR #220, opened 2021-02-17
feat: added cascade option to DeleteResource - argo-cd #5368

ishitasequeira:
Signed-off-by: Ishita Sequeira isequeir@redhat.com
issue: argoproj/argo-cd#[5368](https://github.com/argoproj/argo-cd/issues/5368)

Added cascade option to DeleteResource - argo-cd #5368

### DIFF
--- a/pkg/utils/kube/ctl.go
+++ b/pkg/utils/kube/ctl.go
@@ -186,7 +186,7 @@ func (k *KubectlCmd) PatchResource(ctx context.Context, config *rest.Config, gvk
 	return resourceIf.Patch(ctx, name, patchType, patchBytes, metav1.PatchOptions{}, subresources...)
 }
 
-// DeleteResource deletes resource
+// DeleteResource deletes resource by 3 ways foreground, background and orphan
 func (k *KubectlCmd) DeleteResource(ctx context.Context, config *rest.Config, gvk schema.GroupVersionKind, name string, namespace string, forceDelete bool, cascadeDelete bool) error {
 	span := k.Tracer.StartSpan("DeleteResource")
 	span.SetBaggageItem("kind", gvk.Kind)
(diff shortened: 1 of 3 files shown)


pkg/utils/kube/ctl.go
```diff
@@ -213,7 +213,9 @@ func (k *KubectlCmd) DeleteResource(ctx context.Context, config *rest.Config, gv
 		zeroGracePeriod := int64(0)
 		deleteOptions.GracePeriodSeconds = &zeroGracePeriod
 	}
-
+	if cascadeDelete {
+		propagationPolicy = metav1.DeletePropagationOrphan
+	}
```
jannfis, line 218: I think since the `propagationPolicy` is an either/or (so, either foreground, background or orphaned), me might want to reflect that in the signature of `DeleteResource` method as well. With the current change in this PR, setting `forceDelete` and `cascadeDelete` to true would result in a kind of undefined behavior. Or is my assumption wrong, and `GracePeriodSeconds` should/can also be set for `orphaned` propagation policy?

If my assumption is correct, my proposal is, instead of introducing a new bool flag to the list of parameters (`cascadeDelete`), we should replace `forceDelete bool` with `propagationPolicy metav1.DeletionPropagetion` as the parameter, and rewrite logic in this method, as well as its callers.

jannfis (changes requested):
@ishitasequeira Thanks you for this PR! Please note my review comment and let me know what you think.

pkg/utils/kube/ctl.go
ishitasequeira: @jannfis Thanks for the review. I agree with what you suggested. I have addressed the review comment.

ishitasequeira: Thanks, @jannfis for the review. I agree with your suggestion and have updated the PR with the logic.

jannfis (approved):
LGTM, thank you!

As an additional idea, maybe we can provide the gitops-engine consumers with pre-defined, public delete options as short-cuts?

For example,
```go
var DeleteOptionForeground metav1.DeleteOptions = metav1.DeleteOptions{PropagationPolicy: metav1.DeletePropagationForeground}
// etc ...
```
I'm not sure in which package this would best go, but I guess we'll figure that out.

And it can go in a separate PR as well.

PR #225, opened 2021-02-18
fix: sync should apply Namespaces and CRDs before resources that depend on them

alexmt:

PR fixes issues reproduced in Argo CD (https://github.com/argoproj/argo-cd/issues/5522) and Gitlab agent ([#96](https://gitlab.com/gitlab-org/cluster-integration/gitlab-agent/-/issues/96)) .

During the syncing process, the engine should sync namespaces and CRD before dependent resources ( regardless of sync waves/sync hooks annotations ). This behavior was supposed to be implemented by https://github.com/argoproj/gitops-engine/pull/159 however it was done with a bug: current implementation would work only if the sorting function being executed against all possible sync tasks pair which is not true.

PR rollback changes in sorting function and implements the Namespace/CRD handling in a separate function.

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -683,7 +683,7 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
 		}
 	}
 
-	sort.Sort(tasks)
+	tasks.Sort()
 
 	return tasks, successful
 }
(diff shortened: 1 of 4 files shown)


mayzhang2000 (approved):
LGTM

ash2k (approved):
I don't fully understand the code, but the logic and tests looks fine to me.

PR #234, opened 2021-03-10
fix: Add additional healthy states for HPA

mikebryant:
In the case that you have a HPA with a minimum or maximum replica count set, and the metrics would indicate a need to scale beyond those, this is still an expected and healthy state.

I've got a Deployment & HPA, which doesn't have much load right now, but still has a minimum replicas set to 2 (for HA). This means Argo CD is currently constantly showing it as Progressing, which isn't correct.

Here I'm trying to add the additional healthy states I see.

I wonder though, is checking conditions correct, or should this be checking if the `Status` of the condition is True?
(e.g. is healthy == AbleToScale.Status == True && ScalingActive.Status == True)


alexmt (approved):
I was confused by this behavior. Great explanation and fix. Thank you!

PR #236, opened 2021-03-12
fix: Dry run stuck on pre sync hook

mayzhang2000:

fixes: https://github.com/argoproj/argo-cd/issues/5736

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -1046,15 +1046,13 @@ func (sc *syncContext) processCreateTasks(state runState, tasks syncTasks, dryRu
 				state = failed
 			}
 			if !dryRun || sc.dryRun || result == common.ResultCodeSyncFailed {
-				sc.setResourceResult(t, result, operationPhases[result], message)
-			}
-
-			if sc.dryRun {
-				operationPhase := common.OperationSucceeded
-				if result == common.ResultCodeSyncFailed {
-					operationPhase = common.OperationFailed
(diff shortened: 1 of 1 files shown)


pkg/sync/sync_context.go
```diff
@@ -1048,6 +1048,14 @@ func (sc *syncContext) processCreateTasks(state runState, tasks syncTasks, dryRu
 			if !dryRun || sc.dryRun || result == common.ResultCodeSyncFailed {
 				sc.setResourceResult(t, result, operationPhases[result], message)
 			}
+
+			if sc.dryRun {
```
alexmt, line 1052: Nitpick:

If `sc.dryRun==true` then first `if` at line 1048 should set operation result and then new `if` overrides it. Instead I suggest handle both cases in first if as following:

```golang
if !dryRun || sc.dryRun || result == common.ResultCodeSyncFailed {
	phase := operationPhases[result]
	// no resources are created in dry-run, so running phase means validation was
	// successful and sync operation succeeded
	if sc.dryRun && phase == common.OperationRunning {
		phase = common.OperationSucceeded
	}
	sc.setResourceResult(t, result, phase, message)
}
```

alexmt (changes requested):

pkg/sync/sync_context.go
mayzhang2000: Thank you @alexmt !

alexmt (approved):
LGTM!

PR #238, opened 2021-03-15
chore: Use v0.20.4 kube dependencies

sbose78:
Use v0.20.4 kube dependencies


alexmt (approved):
Thank you! I was about to work on this change as well :)

PR #243, opened 2021-03-16
fix: file extension comparisons are case sensitive

mayzhang2000:


alexmt (approved):
Approved!

PR #245, opened 2021-03-17
fix: sync option applyOutOfSync should not skip pruning

kshamajain99:

fixes https://github.com/argoproj/argo-cd/issues/5790

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -562,7 +562,7 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
 		}
 
 		if sc.applyOutOfSyncOnly {
-			if modified, ok := sc.modificationResult[k]; !modified && ok {
+			if modified, ok := sc.modificationResult[k]; !modified && ok && resource.Target != nil {
 				sc.log.WithValues("resource key", k).V(1).Info("Skipping as resource was not modified")
 				continue
 			}


pkg/sync/sync_context.go
```diff
@@ -563,8 +563,12 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
 
 		if sc.applyOutOfSyncOnly {
 			if modified, ok := sc.modificationResult[k]; !modified && ok {
-				sc.log.WithValues("resource key", k).V(1).Info("Skipping as resource was not modified")
-				continue
+				if sc.prune && resource.Target == nil {
```
alexmt, line 566: I think we should not analyze `sc.prune` here, just assume resource is modified if `resource.Target` or `resource.Live` is nil. This way sync result will have information about resource that have to be pruned but skipped since prune is disabled.

alexmt, line 566: I might be wrong but I think we need to check for `resource.Target == nil || resource.Live == nil`. In both cases resource is not in sync.

alexmt (changes requested):

pkg/sync/sync_context.go
kshamajain99: if `resource.Live == nil` then modified will be true. In that case, this resource won't be skip. Hence, we don't need to check if `resource.Live == nil`.

alexmt (approved):
LGTM

PR #246, opened 2021-03-17
feat: support 'Replace=true' sync option

alexmt:

PR introduces support for `Replace=true` sync option that enables use of `kubectl replace` instead of `kubectl apply` .

### DIFF
--- a/pkg/utils/kube/ctl.go
+++ b/pkg/utils/kube/ctl.go
@@ -13,6 +13,7 @@ import (
 	"golang.org/x/sync/errgroup"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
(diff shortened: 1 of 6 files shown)


alexmt:
Ready for review.

pkg/utils/kube/ctl.go
```diff
@@ -328,6 +318,80 @@ func kubeCmdFactory(kubeconfig, ns string) (cmdutil.Factory, genericclioptions.I
...
+	}
+	resource := gvk.GroupVersion().WithResource(apiResource.Name)
+	resourceIf := ToResourceInterface(dynamicIf, apiResource, resource, namespace)
+
+	createOptions := metav1.CreateOptions{}
+	switch dryRunStrategy {
+	case cmdutil.DryRunClient:
+		createOptions.DryRun = []string{metav1.DryRunAll}
+	case cmdutil.DryRunServer:
+		createOptions.DryRun = []string{metav1.DryRunAll}
```
kshamajain99, line 368: Looks like DryRunClient and DryRunServer does the same thing. If so, then why don't we combine both case into one to make it more clearer? Like
```
case cmdutil.DryRunClient, cmdutil.DryRunServer:
       createOptions.DryRun = []string{metav1.DryRunAll}
```

kshamajain99 (approved):
Except for minor change, LGTM

pkg/utils/kube/ctl.go
alexmt: good point! fixed.

PR #250, opened 2021-03-23
Add a normalizer that makes use of well-known types

sbose78:
- [ ] Tests successful
- [ ] Generation of well-known types
- [ ] Does this break any API compatibility?
- [ ] Does this take care of exposing only those functions which are needed to expose?


pkg/diff/normalizer.go
```diff
@@ -0,0 +1,17 @@
+package diff
+
+import (
+	"github.com/argoproj/gitops-engine/pkg/diff/normalizer/knowntypes"
+	"github.com/argoproj/gitops-engine/pkg/diff/normalizer/noop"
+)
+
+// GetNoopNormalizer returns normalizer that does not apply any resource modifications
+func GetNoopNormalizer() Normalizer {
```
sbose78, line 9: should be non-exported method, unsure if someone is already using it.

Makefile
```diff
@@ -7,7 +7,12 @@ IMAGE_PREFIX=${IMAGE_NAMESPACE}/
 endif
 
 .PHONY: generate
-generate: agent-manifests
+generate: gogen agent-manifests
+
+
+.PHONY: gogen
+gogen:
+	go generate ./pkg/diff/normalizer/knowntypes/...
```
sbose78, line 15: doesn't work yet

pkg/diff/normalizer/knowntypes/normalizer_test.go
```diff
@@ -0,0 +1,265 @@
...
+kind: TestCRD
+metadata:
+  name: canary-demo
+spec:
+  ram: 1.25G`)
+
+	n := &KnownTypesNormalizer{
+		typeFields: map[schema.GroupKind][]KnownTypeField{},
+	}
+	err := n.addKnownField(schema.GroupKind{Group: "some.io", Kind: "TestCRD"}, "spec.ram", "core/Quantity")
```
sbose78, line 200: This is a little diversion from how we do it in ArgoCD

PR #252, opened 2021-03-25
feat: support replace strategy for CRD

alexmt:

Addresses issue reported in https://github.com/argoproj/argo-cd/issues/820 for CRDs. The sync should attempt to just update CRD spec if `Replace=true` sync option is specified. The update won't attempt to recreate CRD so it is safe and it is not trying to set `last-applied-configuration` annotation .


kshamajain99 (approved):
LGTM

travisghansen:
Is this in `v2.0.0`? I'm also slightly confused by the intent here...if I don't want to use `replace` for all the other assets I should be manually selecting the `crds` and just syncing those with this option set? Or should this inherently use the `replace` behavior for all `crds` regardless of what the value is set at for all the other resources?

PR #253, opened 2021-04-05
fix: applyoutofsync with dry-run

kshamajain99:

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -374,8 +374,14 @@ func (sc *syncContext) Sync() {
 		// to perform the sync. we only wish to do this once per operation, performing additional dry-runs
 		// is harmless, but redundant. The indicator we use to detect if we have already performed
 		// the dry-run for this operation, is if the resource or hook list is empty.
-		sc.log.WithValues("tasks", tasks).Info("Tasks (dry-run)")
-		if sc.runTasks(tasks, true) == failed {
+
+		dryRunTasks := tasks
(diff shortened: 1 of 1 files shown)


pkg/sync/sync_context.go
```diff
@@ -513,6 +513,17 @@ func (sc *syncContext) Sync() {
 	}
 }
 
+// filter out out-of-sync tasks
+func (sc *syncContext) filterOutOfSyncTasks(tasks syncTasks) syncTasks {
+	return tasks.Filter(func(t *syncTask) bool {
+		if modified, ok := sc.modificationResult[t.resourceKey()]; !modified && ok && t.targetObj != nil && t.liveObj != nil {
```
alexmt, line 519: Found another issue while testing the change. Sync hooks with hard-coded names should not filtered out. It is really easy to fix:

```golang
func (sc *syncContext) filterOutOfSyncTasks(tasks syncTasks) syncTasks {
	return tasks.Filter(func(t *syncTask) bool {
		if t.isHook() {
			return true
		}
		if modified, ok := sc.modificationResult[t.resourceKey()]; !modified && ok && t.targetObj != nil && t.liveObj != nil {
			sc.log.WithValues("resource key", t.resourceKey()).V(1).Info("Skipping as resource was not modified")
			return false
		}
		return true
	})
}
```

alexmt (changes requested):

alexmt (approved):
LGTM

PR #256, opened 2021-04-17
fix: missing resource version in tests

### LINKED ISSUE #254: Flaky test(s)
Failure in CI: https://github.com/argoproj/gitops-engine/runs/2273876928?check_suite_focus=true

I can reproduce some (different) failures locally with
```
go test  -race -count=100 ./pkg/cache
```
`TestChildDeletedEvent` fails locally for me.

ash2k:
`FakeDynamicClient` returns typed lists (not unstructured lists) since Kubernetes 1.20. The type cast now handles that. It's a bug I introduced in #195, sorry about that.

Fixes #254.


alexmt (approved):
Thank you @ash2k !

PR #258, opened 2021-04-20
refactor: reduce usage of k8s.io/kubernetes packages take 2

ash2k:
Updates #56.

This is a reduced version of https://github.com/argoproj/gitops-engine/pull/205 that does not remove conversion.


alexmt (approved):
LGTM.

PR #281, opened 2021-05-27
refactor: cache cmdutil.Factory to improve synchronization performance

alexmt:

Changes implemented in PR ensures that `sync_context` uses one instance of `cmdutil.Factory` from `github.com/kubernetes/kubectl` package to perform `kubectl apply/create/replace` operations. This significantly improves syncing performance and reduces memory usage. 

Tested with a sample app that consists of ~40 Config Maps. Synchronization is ~100 times faster (~0.2s instead of ~30s) and ~100 less memory usage ( ~16 mb instead of 1.3gb ).


kshamajain99 (approved):
LGTM

PR #288, opened 2021-06-16
fix: use new kubectl cache directory for each sync operation

alexmt:


alexmt:
Cached discovery client creates several megabytes of files per cluster. It uses too much memory.

PR #296, opened 2021-06-23
fix: workflows with empty status.phase should be considered Progressing

jessesuen:
Resolves https://github.com/argoproj/argo-cd/issues/6541

Workflows that have an empty `status.phase` have yet to be observed by the workflow controller, and should be considered Progressing, not Healthy.

Signed-off-by: Jesse Suen <jessesuen@gmail.com>


alexmt (approved):
LGTM

PR #298, opened 2021-06-30
fix: 'ResourceOperations.CreateResource' should use 'kubectl' package to properly execute create operation

alexmt:

Closes https://github.com/argoproj/argo-cd/issues/6578

PR changes implementation of  'ResourceOperations.CreateResource' to use kubectl package so it won't fail incorrectly during dry run.


mayzhang2000 (approved):
LGTM

PR #308, opened 2021-07-29
feat: Add support for resource anti-theft

alexec:
Context:

Yesterday I accidentally deleted someone else's resources. To do this:

1. Make error in my manifests that create a cluster role with the same names as someone else's.
2. Sync my app, stealing the resources from the other app.
3. Correct the error in my app by prefixing the resource names.
4. Sync with prune (to remove the extraneous resources), which deleted the other user's resource.

I know I had to make a few mistakes, but Argo CD could have been prevented by erroring when I tried to sync and change the label.


alexec:
Ready for review

PR #321, opened 2021-09-01
fix: SyncOption Replace=True is broken

alexmt:

Closes https://github.com/argoproj/argo-cd/issues/7088


rbreeze (approved):
LGTM

PR #323, opened 2021-09-17
fix: Add ScalingDisabled healthy state to HPA

maruina:
If you disable the HPA because you set the tracked deployment to 0, the HPA should report `Healthy` why at the moment it's reporting as progressing.

image

Signed-off-by: Matteo Ruina <matteo.ruina@gmail.com>


alexmt (approved):
LGTM. Special thank you for the test!

PR #343, opened 2021-10-28
fix: add ScaleDownLimit as health state for HPA

harikrongali:
fixes https://github.com/argoproj/argo-cd/issues/7561


alexmt (approved):
LGTM. thank you!

PR #345, opened 2021-11-02
fix: Address issue during diff when secret data is nil

leoluz:

Fixes https://github.com/argoproj/argo-cd/issues/5584


alexmt (approved):
LGTM

PR #346, opened 2021-11-03
feat: add support split yaml that return actual yamls

pasha-codefresh:
Part of https://github.com/argoproj/argo-cd/pull/7595


alexmt (approved):
LGTM

PR #347, opened 2021-11-04
fix: add default protocol to subset of ports if it is empty

ahalay:
Signed-off-by: Roman Rudenko <3kmnazapad@gmail.com>

Fixes https://github.com/argoproj/argo-cd/issues/6300
by adding default protocol to subset of ports if it is not defined before sorting


alexmt (approved):
Thank you for the fix @ahalay ! Sorry for long response - we did not notice the PR

PR #348, opened 2021-11-08
feat: allow impersonate configuration

jonasrutishauser:
With this change it would be possible to make a PR in argo-cd to support impersonation when writing to the cluster (argoproj/argo-cd#5275).


mikhail-khodorovskiy:
It's been a couple of years - is the impersonation feature now available via other means?

PR #353, opened 2021-11-23
feat: configurable watch resync timeout. ability to disable cluster resync

jessesuen:
When troubleshooting a users performance problem, I found many tunable were not configurable and hard wired into the gitops engine. So I made the following changes:

* Adds the ability to disable the 24h (12h in Argo CD) cluster resync period by setting the value to 0. This will result in the gitops-engine relying on the resource watch resync to correctly resync the cache.
* Also makes the resource watch resync period configurable (from the default of 10m)
* Watch resource resync can also be disabled by setting a value to 0

Signed-off-by: Jesse Suen <jesse@akuity.io>

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -32,10 +32,14 @@ import (
 )
 
 const (
-	clusterResyncTimeout       = 24 * time.Hour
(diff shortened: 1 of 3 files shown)


alexmt (approved):
LGTM

PR #361, opened 2021-12-22
fix: removeNamespaceAnnotation should not panic if annotation has unexpected value

leoluz:
on behalf of @alexmt 


alexmt (approved):
LGTM

PR #363, opened 2022-01-08
Server-side apply

sathieu:
See https://github.com/argoproj/argo-cd/issues/2267

### DIFF
--- a/pkg/sync/common/types.go
+++ b/pkg/sync/common/types.go
@@ -27,6 +27,8 @@ const (
 	SyncOptionPruneLast = "PruneLast=true"
 	// Sync option that enables use of replace or create command instead of apply
 	SyncOptionReplace = "Replace=true"
+	// Sync option that enables use of --server-side flag instead of client-side
+	SyncOptionServerSideApply = "ServerSideApply=true"
 )
 
 type PermissionValidator func(un *unstructured.Unstructured, res *metav1.APIResource) error
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -179,6 +179,12 @@ func WithReplace(replace bool) SyncOpt {
 	}
 }
 
+func WithServerSideApply(serverSideApply bool) SyncOpt {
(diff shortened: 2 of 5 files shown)


alexmt (approved):
LGTM! Thank you @sathieu !


@leoluz can you please take a look as well? I think you know more about server side apply than me.

leoluz:
When using server-side apply it is important to define the `fieldManager` to set the fields ownership to the proper actor. This is a [required field](https://github.com/kubernetes/apimachinery/blob/v0.22.2/pkg/apis/meta/v1/types.go#L571-L579) for `application/apply-patch` patches which is the [one used by SSA](https://github.com/kubernetes/kubectl/blob/v0.22.2/pkg/cmd/apply/apply.go#L443).

Im not sure if this the behaviour using the go client but in kubectl CLI if we don't provide the `fieldManager` it will automatically assign the manager as "kubectl" which is also undesirable.

I believe that we should expose a configuration to also set the manager when choosing server-side apply.

jannfis:
> I believe that we should expose a configuration to also set the manager when choosing server-side apply.

I agree. Ideally, multiple consumers of gitops-engine (e.g. Argo CD instances) would be able to manage different parts of the same resource as well, as this is one of the main advantages of SSA. For Argo CD, we would need a unique identifier (maybe `argocd-<installation_namespace>` as the manager.

Also I wonder about what happens with the `applicationInstance` label, currently used to identify (exclusively) managed resources. Will Argo CD still set that label on a resource that is managed using Server-Side-Apply? How could we have multiple instances of Argo CD manage different parts of the same resource without them overwriting the label, or triggering a shared resource warning?

varunsatokar:
When is this feature planned for release?

leoluz:
@varunsatokar We are currently discussing the proposal and analysing the impact it will bring.
You are welcome to participate in the review if you have something to add:
https://github.com/argoproj/argo-cd/pull/8812

leoluz:
FYI, just rebased this branch with `master` to address conflict.

Issue #214, opened 2021-02-07
Health check for HPA doesn't catch all good states


jannfis:
The health check for HPA doesn't seem to catch all the states which we could consider as healthy.

For example, given the following status, HPA should be in healthy instead of progressing state:

```yaml
    - lastTransitionTime: '2020-09-02T21:38:22Z'
      message: recommended size matches current size
      reason: ReadyForNewScale
      status: 'True'
      type: AbleToScale
```

I think when checking for the health conditions [here](https://github.com/argoproj/gitops-engine/blob/c5b7114c501fc0a12a12c9055ed0913be7c278b3/pkg/health/health_hpa.go#L144-L147), we should rather check for `status` to be `True` instead of matching the `reason` field against known patterns.

shaoxt:
@alexmt @harikrongali

PR #369, opened 2022-01-24
fix: health check for HPA doesn't catch all good states

alexmt:

Closes https://github.com/argoproj/gitops-engine/issues/214

HPA health assessment function should check only condition type and status.


mayzhang2000 (approved):
LGTM

PR #373, opened 2022-02-07
feat: expose cluster sync retry timeout

yeya24:
Signed-off-by: Ben Ye <ben.ye@bytedance.com>

This pr exposes the sync retry duration when there is a sync error. The default is only 10s and for some large clusters, several unknown state apps will trigger a lot of full cluster sync every 10s.
Making it configurable improves the performance in the case above. Once this pr gets accepted then I will update the ArgoCD side as well.


yeya24:
cc @alexmt @yydzhou

alexmt (approved):
LGTM. Thank you @yeya24

jannfis:
Cherry-picked this commit into `release-0.5` branch

PR #377, opened 2022-02-12
feat: add IterateGroupHierarchy method to cache

yeya24:
Signed-off-by: Ben Ye <ben.ye@bytedance.com>

This pr adds a IterateGroupHierarchy method to execute multiple IterateHierarchy calls using goroutines and only takes the cluster read lock once.

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -100,6 +100,8 @@ type ClusterCache interface {
 	FindResources(namespace string, predicates ...func(r *Resource) bool) map[kube.ResourceKey]*Resource
 	// IterateHierarchy iterates resource tree starting from the specified top level resource and executes callback for each resource in the tree
 	IterateHierarchy(key kube.ResourceKey, action func(resource *Resource, namespaceResources map[kube.ResourceKey]*Resource))
+	// IterateGroupHierarchy executes multiple IterateHierarchy operations and only needs to take lock once.
+	IterateGroupHierarchy(key []kube.ResourceKey, action func(resource *Resource, namespaceResources map[kube.ResourceKey]*Resource))
 	// IsNamespaced answers if specified group/kind is a namespaced resource API or not
 	IsNamespaced(gk schema.GroupKind) (bool, error)
 	// GetManagedLiveObjs helps finding matching live K8S resources for a given resources list.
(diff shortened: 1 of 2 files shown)


pkg/cache/cluster.go
```diff
@@ -813,6 +815,53 @@ func (c *clusterCache) IterateHierarchy(key kube.ResourceKey, action func(resour
...
+// IterateGroupHierarchy executes multiple IterateHierarchy operations and only takes lock once.
+func (c *clusterCache) IterateGroupHierarchy(keys []kube.ResourceKey, action func(resource *Resource, namespaceResources map[kube.ResourceKey]*Resource)) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	var wg sync.WaitGroup
+	for _, key := range keys {
+		if res, ok := c.resources[key]; ok {
+			wg.Add(1)
+			go func() {
```
alexmt, line 827: Running separate goroutines for each resource is pretty expensive. I've tried to test this change on a large argocd instance and controller crashed.

pkg/cache/cluster.go
```diff
@@ -813,6 +815,53 @@ func (c *clusterCache) IterateHierarchy(key kube.ResourceKey, action func(resour
 	}
 }
 
+// IterateGroupHierarchy executes multiple IterateHierarchy operations and only takes lock once.
+func (c *clusterCache) IterateGroupHierarchy(keys []kube.ResourceKey, action func(resource *Resource, namespaceResources map[kube.ResourceKey]*Resource)) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	var wg sync.WaitGroup
+	for _, key := range keys {
```
alexmt, line 824: Instead of copying whole `IterateHierarchy` method it is better to create private `iterateHierarchy` method that does not use lock and use it in both `IterateGroupHierarchy` and `IterateHierarchy`

alexmt (changes requested):
Thank you @yeya24 . I've added few comments to the PR. Tried to resolve it in my own [fork](https://github.com/alexmt/argo-cd/tree/add-iterate-group-hierarchy) and test on internal Argo CD instance, but unfortunately did not get any visible performance improvement.


I think we should not introduce an additional complexity without performance benefits. I'm trying to implement another approach we've discussed offline: introduce resource -> children index and remove scan from `IterateHierarchy` method.

yeya24:
Thanks @alexmt. How is your setup and scale for the test instance?
In our case, this did help a little bit when lock contention is huge. This change will only acquire the lock once instead of once for every `IterateHierachy` call so it reduces the probability of waiting for the read lock for a long time.

But if your environment doesn't have the cluster lock contention situation, then I think it is expected there is not much performance improvement.

alexmt:
@yeya24 , I was testing an argocd that manages lot of clusters but each cluster is relatively small. Maybe we need a large cluster with multiple apps to notice a performance improvement. I've built Argo CD image using this change https://github.com/alexmt/gitops-engine/commit/f8ab9ddca97f2ea1734a631c4b58661d1f9e9f92 . Do you want to test this change in your environment? Let me know please - I can build an image using v2.3 release + changes from https://github.com/alexmt/gitops-engine/commit/f8ab9ddca97f2ea1734a631c4b58661d1f9e9f92

yeya24:
Unfortuantely I cannot test this in our environment right now as we fixed the bug mentioned in https://github.com/argoproj/argo-cd/issues/8428 so there is almost no lock contention right now.

But when that bug existed this change did help us reduce some P99 time from like from 2m to 1m to improve the read lock. But it doesn't help much for the normal case so you cannot tell the difference when you didn't hit that issue.

PR #381, opened 2022-02-16
chore: Remove support for deprecated extensions APIs

terrytangyuan:
This PR removed all extensions APIs besides Ingress because extensions Ingress is still available in 1.21. We will remove it later when 1.21 has reached to EOL.

Reference: https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

Signed-off-by: Yuan Tang <terrytangyuan@gmail.com>

### DIFF
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -111,17 +111,6 @@ func GetHealthCheckFunc(gvk schema.GroupVersionKind) func(obj *unstructured.Unst
 		case kube.DaemonSetKind:
 			return getDaemonSetHealth
 		}
-	case "extensions":
-		switch gvk.Kind {
-		case kube.DeploymentKind:
-			return getDeploymentHealth
-		case kube.IngressKind:
-			return getIngressHealth
-		case kube.ReplicaSetKind:
-			return getReplicaSetHealth
-		case kube.DaemonSetKind:
-			return getDaemonSetHealth
-		}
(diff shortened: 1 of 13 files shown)


pkg/health/health.go
```diff
@@ -111,17 +111,6 @@ func GetHealthCheckFunc(gvk schema.GroupVersionKind) func(obj *unstructured.Unst
 		case kube.DaemonSetKind:
 			return getDaemonSetHealth
 		}
-	case "extensions":
-		switch gvk.Kind {
-		case kube.DeploymentKind:
-			return getDeploymentHealth
-		case kube.IngressKind:
```
terrytangyuan, line 118: TODO: Need to add this back since extensions Ingress is still available/served in K8s 1.21 
> The v1.22 release will stop serving the following deprecated API versions in favor of newer and more stable API versions:

> Ingress in the extensions/v1beta1 API version will no longer be served
        Migrate to use the networking.k8s.io/v1beta1 API version, available since v1.14. Existing persisted data can be retrieved/updated via the new version.

terrytangyuan:
Marking as draft to wait until Argo CD 2.3 is released to avoid any risks

terrytangyuan:
Now that 2.3.0 has been released. This is ready for review. @jessesuen @alexmt

alexmt (approved):
LGTM

PR #388, opened 2022-03-15
feat: support exiting early from IterateHierarchy method

alexmt:


pkg/cache/cluster.go
```diff
@@ -101,7 +101,7 @@ type ClusterCache interface {
 	// FindResources returns resources that matches given list of predicates from specified namespace or everywhere if specified namespace is empty
 	FindResources(namespace string, predicates ...func(r *Resource) bool) map[kube.ResourceKey]*Resource
 	// IterateHierarchy iterates resource tree starting from the specified top level resource and executes callback for each resource in the tree
-	IterateHierarchy(key kube.ResourceKey, action func(resource *Resource, namespaceResources map[kube.ResourceKey]*Resource))
+	IterateHierarchy(key kube.ResourceKey, action func(resource *Resource, namespaceResources map[kube.ResourceKey]*Resource) bool)
```
jessesuen, line 104: Can you document what the bool return value of the action callback is supposed to indicate?

pkg/cache/cluster.go
```diff
@@ -846,14 +846,15 @@ func (c *clusterCache) IterateHierarchy(key kube.ResourceKey, action func(resour
...
 				child := children[0]
-				action(child, nsNodes)
-				child.iterateChildren(nsNodes, map[kube.ResourceKey]bool{res.ResourceKey(): true}, func(err error, child *Resource, namespaceResources map[kube.ResourceKey]*Resource) {
-					if err != nil {
-						c.log.V(2).Info(err.Error())
-						return
-					}
-					action(child, namespaceResources)
-				})
+				if action(child, nsNodes) {
```
jessesuen, line 849: `action()` is called on line 831, but I don't see the return value being checked. Is that expected?

alexmt: Thank you! This is a bug. Added missing check and unit test

pkg/cache/cluster.go
alexmt: done

alexmt:
Thank you for review @jessesuen , PTAL

jessesuen (approved):
LGTM

PR #395, opened 2022-03-28
fix: Set QPS and burst rate for resource ops client

jannfis:
This allows to set Kubernetes client QPS and burst rates for the resource ops client, according to the configuration of the client.

Needs a very small change in Argo CD as well, which will be submitted separately.

Supersedes #349 


alexmt (approved):

jannfis:
Cherry-picked into release-0.6: https://github.com/argoproj/gitops-engine/commit/1a7a69fa002a1f3ad3b9935f4281c7cb38907c9e

PR #403, opened 2022-04-12
chore: test resource conversion

crenshaw-dev:
Argo CD currently has an e2e test which verifies that, if a resource version conversion isn't possible locally, we'll fall back to server-side conversion.

That test is difficult to maintain because it relies on having a CRD which is 1) not convertible locally and 2) _is_ convertible server-side for all supported k8s versions and whatever version of k8s the developer is running locally.

I want to [remove that e2e test](https://github.com/argoproj/argo-cd/issues/9046) because the maintenance overhead is excessive.

This PR checks for the behavior that the e2e test checks - do we fall back to server-side conversion if local conversion fails.

### DIFF
--- a/pkg/cache/cluster_test.go
+++ b/pkg/cache/cluster_test.go
@@ -490,6 +490,8 @@ metadata:
 }
 
 func TestGetManagedLiveObjsFailedConversion(t *testing.T) {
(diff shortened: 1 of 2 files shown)


leoluz (approved):
LGTM

PR #404, opened 2022-04-20
feat: add GvkParser in cluster cache

leoluz:
Relates to https://github.com/argoproj/argo-cd/issues/9071


crenshaw-dev (approved):
lgtm

rishabh625:
Hey @leoluz : m still trying to understand the code, but i have a question, Was this change imported in argo-cd and tested ? (ValidatingWebhookConfiguration case): Bug causing wrong diff calculation inside nested lists in CRDs with managed fields ignore configs.

rishabh625:
> Hey @leoluz : m still trying to understand the code, but i have a question, Was this change imported in argo-cd and tested ? (ValidatingWebhookConfiguration case): Bug causing wrong diff calculation inside nested lists in CRDs with managed fields ignore configs.
> 
> is there any fork we could know that this fixes above

looks like this has https://github.com/argoproj/argo-cd/pull/9170/files

alexmt (approved):
LGTM

PR #418, opened 2022-06-14
feat: Add server-side apply manager config

leoluz:
This PR introduces one additional configuration to the `SyncContext` allowing the manager to be specified and used in server-side apply syncs.

Relates to https://github.com/argoproj/argo-cd/issues/2267

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -208,6 +208,18 @@ func buildDiffResult(result *typed.TypedValue, live *unstructured.Unstructured) 
 		if err != nil {
 			return nil, fmt.Errorf("error while marshaling merged object: %w", err)
 		}
+
+		// keep order
+		var result map[string]interface{}
+		err = json.Unmarshal(mergedBytes, &result)
+		if err != nil {
+			return nil, err
+		}
+		mergedBytes, err = json.Marshal(result)
+		if err != nil {
+			return nil, err
+		}
+
 	}
 
 	liveBytes, err := json.Marshal(live)
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -122,7 +122,7 @@ func structuredMergeDiff(config, live *unstructured.Unstructured, gvkParser *man
 	if err != nil {
 		return nil, fmt.Errorf("error building typed value from config resource: %w", err)
 	}
-	tvResult, err := tvLive.Merge(tvConfig)
+	tvResult, err := tvLive.NormalizeUnionsApply(tvConfig)
 	if err != nil {
 		return nil, fmt.Errorf("strategic merge patch error: %w", err)
 	}
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -192,8 +192,13 @@ func buildDiffResult(result *typed.TypedValue, live *unstructured.Unstructured) 
 		return nil, fmt.Errorf("error converting result typedValue: expected map got %T", ru)
 	}
 	mergedUnstructured := &unstructured.Unstructured{Object: r}
-	kubescheme.Scheme.Default(mergedUnstructured)
-	mergedBytes, err := json.Marshal(mergedUnstructured)
+
+	obj, err := scheme.Scheme.New(mergedUnstructured.GroupVersionKind())
+	if err != nil {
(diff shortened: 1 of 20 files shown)


crenshaw-dev (approved):
lgtm!

pkg/diff/diff.go
```diff
@@ -71,6 +75,19 @@ func Diff(config, live *unstructured.Unstructured, opts ...Option) (*DiffResult,
 		live = remarshal(live, o)
 		Normalize(live, opts...)
 	}
+	// structuredMergeDiff is mainly used as a feature flag to enable
+	// calculating diffs using the structured-merge-diff library
+	// used in k8s while performing server-side applies. It checks the
+	// given diff Option or if the desired state resource has the
+	// Server-Side apply sync option annotation enabled.
+	structuredMergeDiff := o.structuredMergeDiff || resource.HasAnnotationOption(config, "argocd.argoproj.io/sync-options", "ServerSideApply=true")
```
leoluz, line 83: I tried to use `common.AnnotationSyncOptions` and `common.SyncOptionServerSideApply` instead of hardcoding the constants here. Unfortunately there is a (hard to solve) cyclic dependency with the `kube` package  if a do so. Im leaving it hardcoded for now but we should maybe refactor the `common` package to remove the `kube` dependency from it.

pkg/utils/kube/scheme/typed.go
```diff
@@ -0,0 +1,72 @@
...
+		return parser.Type(gvk)
+	}
+	return pt
+}
+
+func resolveFromStaticParser(gvk schema.GroupVersionKind, parser *managedfields.GvkParser) *typed.ParseableType {
+	gvkNameMap := getGvkMap(parser)
+	name := gvkNameMap[gvk]
+
+	p := StaticParser()
```
crenshaw-dev, line 32: It's definitely a micro-optimization, but could we avoid this if `name == ""` by returning nil?

crenshaw-dev (approved):
Nitpicks only. lgtm

pkg/diff/diff.go
```diff
@@ -224,6 +368,38 @@ func applyPatch(liveBytes []byte, patchBytes []byte, newVersionedObject func() (
...
+	// so that they can be lexicographically compared with one another).
+	var result map[string]interface{}
+	err = json.Unmarshal([]byte(patchedBytes), &result)
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshaling patched bytes: %w", err)
+	}
+	patchedBytes, err = json.Marshal(result)
+	if err != nil {
+		return nil, fmt.Errorf("error marshaling patched bytes: %w", err)
+	}
```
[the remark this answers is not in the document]
leoluz: Unfortunately it doesn't. See that the standard patch also has the same logic:
https://github.com/argoproj/gitops-engine/blob/a956f804beff3e89a6c1dcd15852b5d52a0858f0/pkg/diff/diff.go#L382-L399

I was having issues with messed diffs before doing this.

pkg/utils/kube/scheme/typed.go
leoluz: sorry but not sure I understand the suggestion. Returning `nil` from where?

crenshaw-dev: Ff name == nil on line 33, we return nil on line 34 and don’t use the output of StaticParser from line 32. We could check name before calling StaticParser and save some time/memory.

pkg/diff/diff.go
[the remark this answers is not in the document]
crenshaw-dev: That's really surprising... The function we're calling ends with a call to `json.Marshal`. The `json.Marshal` technically points to their `util/json` package, but it's just a wrapper for `encoding/json`.
https://github.com/kubernetes/apimachinery/blob/v0.23.1/pkg/util/strategicpatch/patch.go#L812-L837

rafilkmp3 (approved):

rafilkmp3 (approved):

pkg/diff/diff.go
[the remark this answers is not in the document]
leoluz: In fact I removed the previous reordering code and can confirm that it still works as expected. Not sure what cause that code to be necessary in the past. Good catch.

PR #423, opened 2022-06-24
fix: Only consider resources which supports appropriate verb for any given operation

### LINKED ISSUE #422: gitops-engine should consider supported verbs when resolving GVK to resources
Some API extensions provide two or more resources for the same GVK, but for different verbs.

One example of this is the Template API in OpenShift, which is implemented as an API extension instead of a CRD. For example:

```
$ oc api-resources -o wide | grep ' Template '
processedtemplates                                        template.openshift.io/v1                      true         Template                             [create]
templates                                                 template.openshift.io/v1                      true         Template                             [create delete deletecollection get list patch update watch]
```

So the GVK `G=template.openshift.io, K=Template, V=v1` resolves to two APIs, `processedtemplates` and `templates`. However, only the `templates` API supports a full list of verbs such as get, delete, etc.

When performing the conversion from GVK to API resource, gitops-engine just picks the first match, which in above case would result in picking `processedtemplates`. However, gitops-engine should ensure that it also checks whether the resource supports the requested verb for the operation.

One could argue that above example is bad design, and that one GVK should only resolve to a single API and I would agree with that.  I think this quirk is required for some backwards compatibility in OpenShift. However, it is currently not possible to manage Template resources in OpenShift due to this behavior. A Template resource can be created (because OpenShift under the hood converts processedtemplate to template), but processedtemplate only supports the create verb.

jannfis:
This will consider the verbs supported by the API any GVK resolves to.

Fixes #422 

### DIFF
--- a/pkg/utils/kube/kube.go
+++ b/pkg/utils/kube/kube.go
@@ -179,16 +179,24 @@ func IsCRD(obj *unstructured.Unstructured) bool {
 //
 // See: https://github.com/ksonnet/ksonnet/blob/master/utils/client.go
 func ServerResourceForGroupVersionKind(disco discovery.DiscoveryInterface, gvk schema.GroupVersionKind, verb string) (*metav1.APIResource, error) {
+	// default is to return a not found for the requested resource
+	retErr := apierr.NewNotFound(schema.GroupResource{Group: gvk.Group, Resource: gvk.Kind}, "")
 	resources, err := disco.ServerResourcesForGroupVersion(gvk.GroupVersion().String())
 	if err != nil {
 		return nil, err
 	}
 	for _, r := range resources.APIResources {
-		if r.Kind == gvk.Kind && isSupportedVerb(&r, verb) {
-			return &r, nil
+		if r.Kind == gvk.Kind {
+			if isSupportedVerb(&r, verb) {
+				return &r, nil
+			} else {
+				// We have a match, but the API does not support the action
+				// that was requested. Memorize this.
+				retErr = apierr.NewMethodNotSupported(schema.GroupResource{Group: gvk.Group, Resource: gvk.Kind}, verb)
+			}
 		}
 	}
-	return nil, apierr.NewNotFound(schema.GroupResource{Group: gvk.Group, Resource: gvk.Kind}, "")
+	return nil, retErr
 }
 
 var (
--- a/pkg/utils/kube/kube_test.go
+++ b/pkg/utils/kube/kube_test.go
@@ -7,8 +7,13 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	extv1beta1 "k8s.io/api/extensions/v1beta1"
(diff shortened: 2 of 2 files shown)


pkg/utils/kube/kube.go
```diff
@@ -172,14 +172,19 @@ func IsCRD(obj *unstructured.Unstructured) bool {
...
 // See: https://github.com/ksonnet/ksonnet/blob/master/utils/client.go
-func ServerResourceForGroupVersionKind(disco discovery.DiscoveryInterface, gvk schema.GroupVersionKind) (*metav1.APIResource, error) {
+func ServerResourceForGroupVersionKind(disco discovery.DiscoveryInterface, gvk schema.GroupVersionKind, verb string) (*metav1.APIResource, error) {
 	resources, err := disco.ServerResourcesForGroupVersion(gvk.GroupVersion().String())
 	if err != nil {
 		return nil, err
 	}
 	for _, r := range resources.APIResources {
-		if r.Kind == gvk.Kind {
+		if r.Kind == gvk.Kind && isSupportedVerb(&r, verb) {
```
jessesuen, line 187: I'm not sure if users will ever run into a misleading message where we report `NotFound` error in logs but it's not because the GVK doesn't exist, but rather the verb isn't supported and we did not distinguish between the two. Since it's a pretty rare corner case, I don't think it's necessary but wanted to point it out.

jessesuen (approved):

pkg/utils/kube/ctl.go
```diff
@@ -102,10 +103,14 @@ func (k *KubectlCmd) filterAPIResources(config *rest.Config, preferred bool, res
 	return apiResIfs, nil
 }
 
-// isSupportedVerb returns whether or not a APIResource supports a specific verb
+// isSupportedVerb returns whether or not a APIResource supports a specific verb.
+// The verb will be matched case-insensitive.
 func isSupportedVerb(apiResource *metav1.APIResource, verb string) bool {
+	if verb == "" || verb == "*" {
```
crenshaw-dev, line 109: Is this something that can happen? As far as I can tell, all instances of the argument are hard-coded as values others than these two.

crenshaw-dev (approved):
Was puzzled about one thing, but lgtm.

pkg/utils/kube/ctl.go
jannfis: Good question. This is intended to allow the caller to just say "I don't care about the support verbs, just resolve it". It might be a valid use-case for some.

pkg/utils/kube/kube.go
jannfis: I actually agree with this comment, @jessesuen. Edge case or not, I'm going to fix this.

jannfis: Changed the logic to return `MethodNotSupported` if the GVK resolves to a resource which does not support the given verb.

Also, added some unit tests around all that.

jannfis:
I have cherry-picked this change into release-0.7

I want to pull this change into an Argo CD v2.4.z release after the current code freeze period.

PR #429, opened 2022-07-12
Use K8s version with .GitVersion

zvlb:
Hello.
Everywhere in Kubernetes used GitVersion for demonstrated K8s version.
(For example `kubectl get nodes` returned exactly GitVersion of kubernetes)

It's important for me, bc i use ArgoCD for all Kubernetes Clusters and i want see Major.Minor.Patch version of kubernetes. But now i see only Major.Minor version.

### DIFF
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,3 +1,0 @@
-[user]
-        name = Zemtsov Vladimir
-        email = vl.zemtsov@gmail.com
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,0 +1,3 @@
+[user]
+        name = Zemtsov Vladimir
+        email = vl.zemtsov@gmail.com
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/managedfields"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/version"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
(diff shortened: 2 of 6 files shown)


vzemtsov:
If this method used in many project, or need backward compatibility, i can add new method like `GetFullServerVersion` and don't change `GetServerVersion`

vzemtsov:
@alexmt @jannfis @jessesuen 
 Guys?

jannfis:
Hey, sorry @vzemtsov for coming back so late to this PR!

I would suggest making this parameterizeable, e.g. add a new boolean field `fullServerVersion` here: https://github.com/argoproj/gitops-engine/blob/517c1fff4e2d5c70f0af05a98b43ead92742fba5/pkg/cache/cluster.go#L167

Then add an option to toggle it in https://github.com/argoproj/gitops-engine/blob/master/pkg/cache/settings.go

Then, modify https://github.com/argoproj/gitops-engine/blob/517c1fff4e2d5c70f0af05a98b43ead92742fba5/pkg/utils/kube/ctl.go#L299 to return the `*version.Info` instead of a `string` and finally modify https://github.com/argoproj/gitops-engine/blob/517c1fff4e2d5c70f0af05a98b43ead92742fba5/pkg/cache/cluster.go#L283 to return the value appropriate to the `fullServerVersion`

This would give people an option and is backwards compatible in the API.

WDYT?

vzemtsov:
I add changes what you say.
It's looks good, but ArgoCD for get Kubernetes Version use only [this](https://github.com/argoproj/gitops-engine/blob/517c1fff4e2d5c70f0af05a98b43ead92742fba5/pkg/utils/kube/ctl.go#L299) function 

Can you tag this changes? I want add changes to ArgoCD, to make work correctly

vzemtsov:
Oh. Need some time for testing. I'm back later

vzemtsov:
@jannfis Do you can check?

vzemtsov:
@jannfis Hi. Can we finish this?

zvlb:
anyone?

Issue #425, opened 2022-07-04
error="error creating gvk parser: duplicate entry for /v1, Kind=APIGroup" 


dllegru:
After a recent upgrade to argo-cd `v2.4.3` coming from `v2.2.x` we're seeing the following error in `argocd-application-controller` component:

```
time="2022-07-04T16:18:16Z" level=error msg="warning loading openapi schema" error="error creating gvk parser: duplicate entry for /v1, Kind=APIGroup" server="https://172.16.8.2"
time="2022-07-04T16:18:17Z" level=error msg="warning loading openapi schema" error="error creating gvk parser: duplicate entry for /v1, Kind=APIGroup" server="https://172.16.8.2"
time="2022-07-04T16:18:25Z" level=error msg="warning loading openapi schema" error="error creating gvk parser: duplicate entry for /v1, Kind=APIGroup" server="https://kubernetes.default.svc"
time="2022-07-04T16:18:25Z" level=error msg="warning loading openapi schema" error="error creating gvk parser: duplicate entry for /v1, Kind=APIGroup" server="https://kubernetes.default.svc"
time="2022-07-04T16:18:36Z" level=error msg="warning loading openapi schema" error="error creating gvk parser: duplicate entry for /v1, Kind=APIGroup" server="https://172.16.8.2"
time="2022-07-04T16:18:36Z" level=error msg="warning loading openapi schema" error="error creating gvk parser: duplicate entry for /v1, Kind=APIGroup" server="https://172.16.8.2"
time="2022-07-04T16:18:41Z" level=error msg="warning loading openapi schema" error="error creating gvk parser: duplicate entry for /v1, Kind=APIGroup" server="https://172.16.8.2"
time="2022-07-04T16:18:43Z" level=error msg="warning loading openapi schema" error="error creating gvk parser: duplicate entry for /v1, Kind=APIGroup" server="https://172.16.8.2"
time="2022-07-04T16:18:44Z" level=error msg="warning loading openapi schema" error="error creating gvk parser: duplicate entry for /v1, Kind=APIGroup" server="https://172.16.8.2"
```

The error is constantly outputed, as you can see it is being reported for both clusters we have under `argo-cd` management.

I guess it is very related to this last [commit](https://github.com/argoproj/gitops-engine/commit/f9456de217f8fb8893d09998bab0d8efff95f4fd) from release [v0.7.0](https://github.com/argoproj/gitops-engine/releases/tag/v0.7.0) @leoluz 

[...]

PR #430, opened 2022-07-13
fix: Change wrong log level in cluster.go openAPISchema, gvkParser

dllegru:
Fixes argoproj/gitops-engine#425

Change wrong log level from `Error` to `Warning`.

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -666,7 +666,7 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
 							if !ok {
 								return err
 							}
-							c.log.Error(e, "warning loading openapi schema")
+							c.log.Warning(e, "warning loading openapi schema")
 						}
 						if gvkParser != nil {
 							c.gvkParser = gvkParser
@@ -730,7 +730,7 @@ func (c *clusterCache) sync() error {
 		if !ok {
 			return err
 		}
-		c.log.Error(e, "warning loading openapi schema")
+		c.log.Warning(e, "warning loading openapi schema")
 	}
 
 	if gvkParser != nil {
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -666,7 +666,7 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
 							if !ok {
 								return err
 							}
-							c.log.Info(e, "warning loading openapi schema")
+							c.log.Info("warning loading openapi schema: %s", e)
 						}
 						if gvkParser != nil {
 							c.gvkParser = gvkParser
@@ -730,7 +730,7 @@ func (c *clusterCache) sync() error {
 		if !ok {
 			return err
 		}
-		c.log.Info(e, "warning loading openapi schema")
+		c.log.Info("warning loading openapi schema: %s", e)
 	}
 
 	if gvkParser != nil {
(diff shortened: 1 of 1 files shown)


leoluz:
@dllegru The reason why I added the log as `Error` is because the log interface doesn't have the `Warning` method. This is why your PR is failing with error:

```
pkg/cache/cluster.go:733: c.log.Warning undefined (type logr.Logger has no field or method Warning)
```

dllegru:
> @dllegru The reason why I added the log as `Error` is because the log interface doesn't have the `Warning` method. This is why your PR is failing with error:
> 
> ```
> pkg/cache/cluster.go:733: c.log.Warning undefined (type logr.Logger has no field or method Warning)
> ```
> 
> We need to define and implement this method. Would you be willing to go ahead and update your PR with this implementation?

I'm sorry @leoluz but my current experience with Golang at the moment is close to 0, so can't help much at this stage.

leoluz:
@dllegru No problem.. I dug a bit and realized that the problem is actually caused by the library we use that [doesn't provide a log.Warning](https://github.com/go-logr/logr/blob/b359493a99cc734afcafd0dec505149d07f78ae1/README.md#why-not-named-levels-like-infowarningerror) method.
If you can just change the log to `Info` in your PR it will work and it should be fine to merge. Also, if you are updating your  PR pls don't forget to signoff your commits (git commit -sm 'commit message') or the DCO check will fail.
Thank you.

dllegru:
Hi @leoluz, thanks for digging into it and sharing the documentation, it's very insightftul.
I've changed to `c.log.Info` as suggested although the tests now fail with:
```
Error: pkg/cache/cluster.go:669:18: cannot use e (type *kube.CreateGVKParserError) as type string in argument to c.log.Info
Error: pkg/cache/cluster.go:733:13: cannot use e (type *kube.CreateGVKParserError) as type string in argument to c.log.Info
```
It seems the problem is in using string format var `e`  cannot be used with `c.log.Info`. I'm sorry can't provide much more help at this moment due my lack of knowledge with Go.

pkg/cache/cluster.go
```diff
@@ -666,7 +666,7 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
 							if !ok {
 								return err
 							}
-							c.log.Error(e, "warning loading openapi schema")
+							c.log.Info(e, "warning loading openapi schema")
```
leoluz, line 669: Please replace this line with:
`c.log.Info("warning loading openapi schema: %s", e)`

pkg/cache/cluster.go
```diff
@@ -730,7 +730,7 @@ func (c *clusterCache) sync() error {
 		if !ok {
 			return err
 		}
-		c.log.Error(e, "warning loading openapi schema")
+		c.log.Info(e, "warning loading openapi schema")
```
leoluz, line 733: Please replace this line with:
`c.log.Info("warning loading openapi schema: %s", e)`

leoluz (changes requested):
Please take a look at the change suggestions to fix the build.

leoluz:
CI/Test job is now passing. Thank you.
You still need to signoff your commits for me the approve/merge. You can do so by running the following commands in your terminal:

```
git rebase HEAD~3 --signoff
git push --force-with-lease origin 425_error_creating_gvk_parser
```

dllegru:
> CI/Test job is now passing. Thank you. @dllegru You still need to signoff your commits for me to approve/merge. You can do so by running the following commands in your terminal:
> 
> ```
> git rebase HEAD~3 --signoff
> git push --force-with-lease origin 425_error_creating_gvk_parser
> ```

Many thanks to you @leoluz! 
I've rebased all the commits, as you pointed I had a missing sign-off message (thanks for the command) in a previous commit and also the email address was wrongly set-up.

I think finally all looks good now! 🥳

leoluz (approved):
LGTM

PR #435, opened 2022-08-03
fix: structured-merge diff apply default values in live resource

leoluz:

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -184,7 +184,7 @@ func structuredMergeDiff(config, live *unstructured.Unstructured, pt *typed.Pars
 		return nil, fmt.Errorf("error applying default values in live: %w", err)
 	}
 
-	return buildDiffResult(predictedLive, taintedLive)
+	return buildDiffResult(predictedLive, taintedLive), nil
 }
 
 func applyDefaultValues(result *typed.TypedValue) ([]byte, error) {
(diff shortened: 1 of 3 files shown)


pkg/diff/diff.go
```diff
@@ -207,16 +213,12 @@ func applyDefaultValues(result *typed.TypedValue) ([]byte, error) {
 	return mergedBytes, nil
 }
 
-func buildDiffResult(mergedBytes []byte, live *unstructured.Unstructured) (*DiffResult, error) {
+func buildDiffResult(predictedBytes []byte, liveBytes []byte) (*DiffResult, error) {
```
crenshaw-dev, line 216: Could this just be incorporated into `structuredMergeDiff`, since it's trivial at this point?

pkg/diff/diff.go
```diff
@@ -207,16 +213,12 @@ func applyDefaultValues(result *typed.TypedValue) ([]byte, error) {
...
 
-	liveBytes, err := json.Marshal(live)
-	if err != nil {
-		return nil, fmt.Errorf("error while marshaling live unstructured: %w", err)
-	}
 	return &DiffResult{
-		Modified:       string(liveBytes) != string(mergedBytes),
+		Modified:       string(liveBytes) != string(predictedBytes),
```
crenshaw-dev, line 219: Is the string conversion necessary for this comparison?

crenshaw-dev (changes requested):
Could the bug be unit tested?

pkg/diff/diff.go
leoluz: get the error:
`operator != not defined for []byte`

pkg/diff/diff.go
leoluz: good idea. done!

leoluz:
@crenshaw-dev improved the unit-test a bit and added a verification to avoid regressions in the current bug

crenshaw-dev (approved):

PR #436, opened 2022-08-04
fix: Fix argument order in resource filter

jannfis:
The signature for the `resourcesFilter` function in the sync context is

```go
resourcesFilter        func(key kube.ResourceKey, target *unstructured.Unstructured, live *unstructured.Unstructured) bool
```

but it's called with `live` and `target` exchanged.

This is a prerequisite for fixing https://github.com/argoproj/argo-cd/issues/8683


ashutosh16 (approved):
LG

alexmt (approved):
LGTM!

PR #440, opened 2022-08-05
feat: add annotation to the custom namesapce

ashutosh16:
Support annotation and labels for custom namespace. 
This is a prerequisite for fixing https://github.com/argoproj/argo-cd/issues/4628

This Argocd  [PR](https://github.com/argoproj/argo-cd/pull/10221) depends on this change
Signed-off-by: Ashutosh <mail.ashutosh8@gmail.com>

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -152,10 +152,11 @@ func WithResourceModificationChecker(enabled bool, diffResults *diff.DiffResultL
 }
 
 // WithNamespaceCreation will create non-exist namespace
-func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool) SyncOpt {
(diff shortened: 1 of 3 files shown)


pkg/sync/sync_context.go
```diff
@@ -152,10 +152,11 @@ func WithResourceModificationChecker(enabled bool, diffResults *diff.DiffResultL
 }
 
 // WithNamespaceCreation will create non-exist namespace
-func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool) SyncOpt {
+func WithNamespaceCreation(createNamespace bool, nsMetaData map[string]interface{}, namespaceModifier func(*unstructured.Unstructured) bool) SyncOpt {
```
crenshaw-dev, line 155: Instead of accepting `map[string]interface{}`, could we just accept two parameters: `labels map[string]string, annotations map[string]string`? That would let us keep all the parsing logic in Argo CD.

crenshaw-dev (changes requested):

ashutosh16:
Not require

PR #443, opened 2022-08-11
feat: Ability to create custom labels for namespaces created with syncOptions CreateNamespace

pasha-codefresh:
https://github.com/argoproj/argo-cd/issues/7799

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -152,10 +152,11 @@ func WithResourceModificationChecker(enabled bool, diffResults *diff.DiffResultL
 }
 
 // WithNamespaceCreation will create non-exist namespace
-func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool) SyncOpt {
+func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool, namespaceCreator func(*unstructured.Unstructured) bool) SyncOpt {
 	return func(ctx *syncContext) {
 		ctx.createNamespace = createNamespace
 		ctx.namespaceModifier = namespaceModifier
+		ctx.namespaceCreator = namespaceCreator
 	}
 }
 
@@ -349,6 +350,7 @@ type syncContext struct {
 
 	createNamespace   bool
 	namespaceModifier func(*unstructured.Unstructured) bool
+	namespaceCreator  func(*unstructured.Unstructured) bool
 
 	syncWaveHook common.SyncWaveHook
 
@@ -792,7 +794,10 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
 					}
 				}
 			} else if apierr.IsNotFound(err) {
-				tasks = append(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj, liveObj: nil})
+				unstructuredObjCopy := unstructuredObj.DeepCopy()
+				if sc.namespaceCreator(unstructuredObjCopy) {
+					tasks = append(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObjCopy, liveObj: nil})
+				}
 			} else {
 				task := &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj}
 				sc.setResourceResult(task, common.ResultCodeSyncFailed, common.OperationError, fmt.Sprintf("Namespace auto creation failed: %s", err))
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -794,9 +794,8 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
 					}
 				}
 			} else if apierr.IsNotFound(err) {
-				unstructuredObjCopy := unstructuredObj.DeepCopy()
-				if sc.namespaceCreator(unstructuredObjCopy) {
-					tasks = append(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObjCopy, liveObj: nil})
+				if sc.namespaceCreator(unstructuredObj) {
+					tasks = append(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj, liveObj: nil})
 				}
 			} else {
 				task := &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj}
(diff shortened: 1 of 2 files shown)


pkg/sync/sync_context.go
```diff
@@ -792,7 +794,10 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
 					}
 				}
 			} else if apierr.IsNotFound(err) {
-				tasks = append(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj, liveObj: nil})
+				unstructuredObjCopy := unstructuredObj.DeepCopy()
+				if sc.namespaceCreator(unstructuredObjCopy) {
+					tasks = append(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObjCopy, liveObj: nil})
+				}
```
crenshaw-dev, line 800: My expectation as a user would be that when I set the `createNamespaceMetadata` field, all auto-created namespaces will be updated with that metadata.

What if in `WithNamespaceCreation` we accepted multiple `namespaceModifier` functions and run them in series where they are currently being run?

crenshaw-dev (changes requested):

pkg/sync/sync_context.go
```diff
@@ -152,10 +152,11 @@ func WithResourceModificationChecker(enabled bool, diffResults *diff.DiffResultL
 }
 
 // WithNamespaceCreation will create non-exist namespace
-func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool) SyncOpt {
+func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool, namespaceCreator func(*unstructured.Unstructured) bool) SyncOpt {
 	return func(ctx *syncContext) {
 		ctx.createNamespace = createNamespace
 		ctx.namespaceModifier = namespaceModifier
+		ctx.namespaceCreator = namespaceCreator
```
ashutosh16, line 159: Do you consider renaming `namespaceCreator`  to `createNamespaceMetadata ` ?  namespaceCreator is very similar to  `createNamespace`

pasha-codefresh: it is hook, inside this hook can be another logic, not only metadata one

pkg/sync/sync_context.go
ashutosh16: @crenshaw-dev  Do you want to add ns metadata to the already created auto namespace?

pasha-codefresh: yes, use case is 

User creates an app with autocreate namespaces
User syncs the app, and namespaces are auto-created
User adds a new label to the createNamespaceMetadata.labels field
User syncs the app, and the label is applied to already-auto-created namespaces


Working on this

pasha-codefresh: @ashutosh16 @crenshaw-dev this use case is covered now 

https://github.com/argoproj/argo-cd/pull/10288/files#diff-627c32a4f31d0f97e6ddc1f19ab3add7fb7791e0d1b0503d78c6c049e8cb4865R260

when the app is going to be synced namespaceModifier will be called and labels will be added to namespace

crenshaw-dev: lgtm!

jannfis, line 800: Is there a good reason for using a copy of `unstructuredObj` here instead of just using the original?

pkg/sync/sync_context_test.go
```diff
@@ -1,9 +1,12 @@
 package sync
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
+	errors2 "k8s.io/apimachinery/pkg/api/errors"
```
jannfis, line 8: Just a nit: Could we please rename the import to `apierrors` or something more obvious than `errors2`?

pkg/sync/sync_context.go
```diff
@@ -152,10 +152,11 @@ func WithResourceModificationChecker(enabled bool, diffResults *diff.DiffResultL
 }
 
 // WithNamespaceCreation will create non-exist namespace
-func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool) SyncOpt {
+func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool, namespaceCreator func(*unstructured.Unstructured) bool) SyncOpt {
```
jannfis, line 155: I think it makes sense to make a unique type of `func (*unstructured.Unstructured) bool`, as the method signature becomes a little unreadable :)

jannfis (commented):
I have a few questions

pkg/sync/sync_context.go
pasha-codefresh: it is in same way for modifier, i decided to do in same way here for prevent issues, probably no 
https://github.com/argoproj/gitops-engine/pull/443/files#diff-74e77a4722803a7d3f87b956ec81b82d3c5a0dd50de299922aa2208fec57e6e9R791

pkg/sync/sync_context_test.go
pasha-codefresh: yes, it was autogenerated, sry

pkg/sync/sync_context.go
pasha-codefresh: Снимок экрана 2022-08-16 в 17 01 44
Not sure that it starts look better, maybe you can advice better naming

pkg/sync/sync_context.go
pasha-codefresh: I think we can remove copy function but in such case we need creator func should apply unstructuredObj  as value and not reference,

pasha-codefresh: yes, it can be removed

jannfis:
Also, looking at the appropriate code in Argo CD, is there a reason we need a new modifier function and do not reuse the existing one here: https://github.com/argoproj/argo-cd/blob/2d544b42c9c41be85a7fcb25535e3013d2ae3350/controller/sync.go#L243? 

This would save us an API change in gitops-engine.

pasha-codefresh:
> 
because modifier call only if namespace already exist, this hook is not calling when namespace not exist
https://github.com/argoproj/gitops-engine/pull/443/files#diff-74e77a4722803a7d3f87b956ec81b82d3c5a0dd50de299922aa2208fec57e6e9R796

@jannfis

jannfis (approved):
Sorry for coming back so late.

LGTM.

PR #448, opened 2022-08-18
fix: cache gvk responses when checking sync task permissions

crenshaw-dev:
It's likely that syncs contain multiple tasks with the same gvk. This will make sure we only hit the k8s API once per gvk per sync operation.


jessesuen:
As part of this change, should we remove caching introduced here:
https://github.com/argoproj/gitops-engine/pull/449/files#diff-74e77a4722803a7d3f87b956ec81b82d3c5a0dd50de299922aa2208fec57e6e9R693-R720

crenshaw-dev:
@jessesuen we should. I'll do that when https://github.com/argoproj/gitops-engine/pull/452 is merged

jessesuen (commented):
Since this PR is attempting to solve our discovery API issues, I'd like to also fix an issue with what we are doing here:

https://github.com/argoproj/gitops-engine/blob/1c4ef336872ce4f30cf2b9134e7f8bcf57c3c813/pkg/utils/kube/resource_ops.go#L141

```golang
	kubeConfigFlags.WithDiscoveryBurst(config.Burst)
	kubeConfigFlags.WithDiscoveryQPS(config.QPS)
```

We uses the same value for client QPS as it does for discovery QPS. This means we are using a value of 50 for discovery. The default value for discovery QPS in client-go v0.25 has been increased from 100->300. So we are effectively hamstringing ourselves.

See: https://github.com/crossplane/crossplane/issues/3272

PR #458, opened 2022-08-30
Feat/enable snapshot counts for

alswl:
Problems:

When I initialize argo controller in  huge cluster(more then 10000 nodes), it will takes about 1 hour to get ready.

I can only observe the progress by `GetClusterInfo`, but this function is blocking by a lock, so I cannot get the verbose progress.

By using this `GetClusterInfoSnapshot`, I can found how many resources is loaded by argocd controller(though this is the exactly).

image


alswl:
ping @terrytangyuan

alswl:
@alexmt Hi, do you have time to look at this Pull Request as well?

pkg/cache/cluster.go
```diff
@@ -1071,6 +1073,19 @@ func (c *clusterCache) GetClusterInfo() ClusterInfo {
 	}
 }
 
+// GetClusterInfoSnapshot returns cluster cache statistics
```
zachaller, line 1076: I would like to see some more details in the comments that these variables are normally held by a lock and that under they can not be modified in this function at all etc. Just to make it more clear that these are normally lock protected.

zachaller:
We should also add docs to https://argo-cd.readthedocs.io/en/stable/operator-manual/metrics/

pkg/cache/cluster.go
alswl: fixed.

alswl:
@zachaller 

> We should also add docs to https://argo-cd.readthedocs.io/en/stable/operator-manual/metrics/

Yes, I created a related pull request(draft) for ArgoCD, It should be replaced by new version gitops-engine after merge.

- discuss: https://github.com/argoproj/argo-cd/issues/11157
- pr: https://github.com/argoproj/argo-cd/pull/11156

alswl:
Friendly ping with @zachaller and @alexmt .

PR #460, opened 2022-09-07
fix: remove last-applied-configuration before diff in ssa

leoluz:
When applying resources with SSA, sometimes last-applied-configuration are present in the live state but not always. This causes the diff to sometimes show the annotation as removed from the desired state. This is incorrect as desired state should never have the last-applied-configuration annotation provided. 

To fix this bug, I am removing the annotation from the live resource as well as from the predicted live so diffs can be displayed correctly.


pkg/diff/diff_test.go
```diff
@@ -786,15 +785,13 @@ func TestStructuredMergeDiff(t *testing.T) {
 		require.NoError(t, err)
 		assert.NotNil(t, result)
 		assert.True(t, result.Modified)
-		assert.Equal(t, string(expectedLiveBytes), string(result.PredictedLive))
+		svc := YamlToSvc(t, result.PredictedLive)
+		assert.Len(t, svc.Spec.Ports, 2)
```
crenshaw-dev, line 789: Why can we no longer do a full text comparison?

leoluz: Basically because it is terrible to maintain. I also found it hard to reason about what the test is actually validating. This way is much cleaner and easier to maintain.

crenshaw-dev (approved):

PR #465, opened 2022-09-22
Update namespace v2

blakepettersson:
Rename `WithNamespaceCreation` to `WithNamespaceModifier`, since this
method is also used for modifying existing namespaces. This method
takes a single argument for the actual updating, and unless this method
gets invoked by its caller no updating will take place (fulfilling what
the `createNamespace` argument used to do).

Within `autoCreateNamespace`, everywhere where we previously added tasks
we'll now need to check whether the namespace should be created (or
modified), which is now delegated to the `appendNsTask` and
`appendFailedNsTask` methods.

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -154,10 +154,11 @@ func WithResourceModificationChecker(enabled bool, diffResults *diff.DiffResultL
 }
 
 // WithNamespaceCreation will create non-exist namespace
-func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool) SyncOpt {
+func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool, namespaceCreator func(*unstructured.Unstructured) bool) SyncOpt {
 	return func(ctx *syncContext) {
 		ctx.createNamespace = createNamespace
 		ctx.namespaceModifier = namespaceModifier
+		ctx.namespaceCreator = namespaceCreator
 	}
 }
 
@@ -351,6 +352,7 @@ type syncContext struct {
 
 	createNamespace   bool
 	namespaceModifier func(*unstructured.Unstructured) bool
+	namespaceCreator  func(*unstructured.Unstructured) bool
 
 	syncWaveHook common.SyncWaveHook
 
@@ -816,7 +818,9 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
 					}
 				}
 			} else if apierr.IsNotFound(err) {
-				tasks = append(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj, liveObj: nil})
+				if sc.namespaceCreator(unstructuredObj) {
+					tasks = append(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj, liveObj: nil})
+				}
 			} else {
 				task := &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj}
 				sc.setResourceResult(task, common.ResultCodeSyncFailed, common.OperationError, fmt.Sprintf("Namespace auto creation failed: %s", err))
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -153,12 +153,13 @@ func WithResourceModificationChecker(enabled bool, diffResults *diff.DiffResultL
 	}
 }
 
-// WithNamespaceCreation will create non-exist namespace
-func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool, namespaceCreator func(*unstructured.Unstructured) bool) SyncOpt {
+// WithNamespaceModifier will create a namespace with the metadata passed in the `*unstructured.Unstructured` argument
+// of the `namespaceModifier` function, in the case it returns `true`. If the namespace already exists, the metadata
+// will overwrite what is already present if `namespaceModifier` returns `true`. If `namespaceModifier` returns `false`,
+// this will be a no-op.
+func WithNamespaceModifier(namespaceModifier func(*unstructured.Unstructured) (bool, error)) SyncOpt {
 	return func(ctx *syncContext) {
-		ctx.createNamespace = createNamespace
 		ctx.namespaceModifier = namespaceModifier
-		ctx.namespaceCreator = namespaceCreator
 	}
 }
 
@@ -350,9 +351,7 @@ type syncContext struct {
 	// lock to protect concurrent updates of the result list
 	lock sync.Mutex
 
-	createNamespace   bool
-	namespaceModifier func(*unstructured.Unstructured) bool
-	namespaceCreator  func(*unstructured.Unstructured) bool
+	namespaceModifier func(*unstructured.Unstructured) (bool, error)
 
 	syncWaveHook common.SyncWaveHook
 
@@ -687,7 +686,7 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
 		}
 	}
 
-	if sc.createNamespace && sc.namespace != "" {
+	if sc.namespaceModifier != nil && sc.namespace != "" {
 		tasks = sc.autoCreateNamespace(tasks)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/sync/sync_context.go
```diff
@@ -153,7 +153,7 @@ func WithResourceModificationChecker(enabled bool, diffResults *diff.DiffResultL
 	}
 }
 
-// WithNamespaceCreation will create non-exist namespace
+// WithNamespaceCreation will create a namespace, if it does not already exist
```
leoluz, line 156: What would happen if the namespace already exists? How will it handle cases when labels/annotations exist in live state with different values than provided in desired state? In this case, it should update the labels/annotations values.

blakepettersson: I clarified the semantics a bit in my latest commit. Do you think that this makes sense (both in wording and semantics)? 😄

pkg/sync/sync_context.go
```diff
@@ -153,7 +153,10 @@ func WithResourceModificationChecker(enabled bool, diffResults *diff.DiffResultL
 	}
 }
 
-// WithNamespaceCreation will create a namespace, if it does not already exist
+// WithNamespaceCreation will create a namespace with the metadata passed in the `*unstructured.Unstructured` argument
+// of the `namespaceModifier` function, in the case it returns `true`. If the namespace already exists, the metadata
+// will overwrite what is already present if `namespaceModifier` returns `true`. If `namespaceModifier` returns `false`,
+// this will be a no-op.
 func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool) SyncOpt {
```
leoluz, line 160: In this case it is better to rename this function to `WithNamespaceModifier`. Just need to be careful to change all references. Also the `createNamespace` parameter seems it isn't making a lot of sense anymore.

leoluz (changes requested):

leoluz:
@blakepettersson Tks for working on this feature!
Make sure you sign off all your commits. The DCO job in our CI needs to be green before we are able to merge your PR.

pkg/sync/sync_context.go
```diff
@@ -153,10 +153,12 @@ func WithResourceModificationChecker(enabled bool, diffResults *diff.DiffResultL
 	}
 }
 
-// WithNamespaceCreation will create non-exist namespace
-func WithNamespaceCreation(createNamespace bool, namespaceModifier func(*unstructured.Unstructured) bool) SyncOpt {
+// WithNamespaceModifier will create a namespace with the metadata passed in the `*unstructured.Unstructured` argument
+// of the `namespaceModifier` function, in the case it returns `true`. If the namespace already exists, the metadata
+// will overwrite what is already present if `namespaceModifier` returns `true`. If `namespaceModifier` returns `false`,
+// this will be a no-op.
+func WithNamespaceModifier(namespaceModifier func(*unstructured.Unstructured) bool) SyncOpt {
```
crenshaw-dev, line 160: Why drop the `createNamespace` param? Seems like it should still be possible for the user to opt entirely out of namespace auto-creation.

blakepettersson:
> Make sure you sign off all your commits. The DCO job in our CI needs to be green before we are able to merge your PR.

It's been well overdue for me to add the git hook, now I've done that 😄

pkg/sync/sync_context.go
blakepettersson: That will still be the case; if I understood @leoluz correctly we can do this within the `namespaceModifier` func itself, and that's how I've currently implemented this in argoproj/argo-cd#7799.

pkg/sync/sync_context.go
blakepettersson: Done!

pkg/sync/sync_context.go
crenshaw-dev: Gotcha. It's a bit bothersome that we have to do a `get` against the k8s API for each ns now rather than skipping the entire process:

```go
liveObj, err := sc.kubectl.GetResource(context.TODO(), sc.config, unstructuredObj.GroupVersionKind(), unstructuredObj.GetName(), metav1.NamespaceNone)
```

But maybe that's not a big deal.

blakepettersson: Ah, that's a good point. 

> But maybe that's not a big deal.

I've no idea myself but it's no biggie to re-add the flag. Saving on k8s API requests sounds like a good thing to me.

blakepettersson: Well we could do something like 

```go
	if sc.namespaceModifier != nil && sc.namespace != "" {
		tasks = sc.autoCreateNamespace(tasks)
	}
```

The calling API feels a bit more awkward though

```go
	var namespaceSync func(un *unstructured.Unstructured) bool
	if syncOp.SyncOptions.HasOption("CreateNamespace=true") {
		namespaceSync = func(un *unstructured.Unstructured) bool {
			return true
		}
	}

# ...
       sync.WithNamespaceModifier(namespaceSync),
```

leoluz: @blakepettersson why not something simpler like this for example?

```golang
	if syncOp.SyncOptions.HasOption("CreateNamespace=true") {
		sync.WithNamespaceModifier(func(un *unstructured.Unstructured) bool {
			return true
		})
	}
// ...

```

blakepettersson: @leoluz  `NewSyncContext` takes all of these options as varargs, so the calling API in Argo currently looks like this 

```go
sync.NewSyncContext(
  compareResult.syncStatus.Revision,
  reconciliationResult,
  restConfig,
  rawConfig,
  m.kubectl,
  app.Spec.Destination.Namespace,
  openAPISchema,
  sync.WithManifestValidation(!syncOp.SyncOptions.HasOption(common.SyncOptionsDisableValidation)),
  sync.WithSyncWaveHook(delayBetweenSyncWaves),
  sync.WithPruneLast(syncOp.SyncOptions.HasOption(common.SyncOptionPruneLast)),
  sync.WithResourceModificationChecker(syncOp.SyncOptions.HasOption("ApplyOutOfSyncOnly=true"), compareResult.diffResultList),
  sync.WithPrunePropagationPolicy(&prunePropagationPolicy),
  sync.WithReplace(syncOp.SyncOptions.HasOption(common.SyncOptionReplace)),
  sync.WithServerSideApply(syncOp.SyncOptions.HasOption(common.SyncOptionServerSideApply)),
  sync.WithServerSideApplyManager(cdcommon.ArgoCDSSAManager),
)
```

We could do something akin to this:

```go
opts := []sync.SyncOpt{ 
     sync.WithLogr(logutils.NewLogrusLogger(logEntry)),
      // All other options
}

if syncOp.SyncOptions.HasOption("CreateNamespace=true") {
    opts = append(opts, sync.WithNamespaceModifier(func(un *unstructured.Unstructured) bool {
	return true
    }))
}

syncCtx, cleanup, err := sync.NewSyncContext(
    compareResult.syncStatus.Revision,
    reconciliationResult,
    restConfig,
    rawConfig,
    m.kubectl,
    app.Spec.Destination.Namespace,
    openAPISchema,
    opts...
)
```

What do you think?

blakepettersson: >  Just to confirm, this way you can avoid the extra call to k8s API (as @crenshaw-dev pointed above) right?

Yes 😄

pkg/sync/sync_context.go
```diff
@@ -811,12 +812,14 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
 				} else {
 					sc.log.WithValues("namespace", sc.namespace).Info("Namespace already exists")
 					liveObjCopy := liveObj.DeepCopy()
-					if sc.namespaceModifier(liveObjCopy) {
+					if liveObj != nil && sc.namespaceModifier(liveObjCopy) {
```
leoluz, line 815: If `liveObj` can be `nil` here, then it is better to move the line above (`liveObjCopy := liveObj.DeepCopy()`) inside the `if` condition to avoid nil pointer errors.

pkg/sync/sync_context.go
leoluz, line 160: the `namespaceModifier` function should probably return error so it can be handled here.

leoluz (changes requested):
Added a few more comments. 
I think we are close to get this PR ready. Tks!

pkg/sync/sync_context.go
blakepettersson: When returning an error from `namespaceModifier`, should we create a new sync task as below?

```go
modified, err := sc.namespaceModifier(liveObjCopy)
if err != nil {
  task := &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj}
  sc.setResourceResult(task, common.ResultCodeSyncFailed, common.OperationError, fmt.Sprintf("Namespace auto creation failed: %s", err))
  tasks = append(tasks, task)
}
```

Or should we set the sync context operation phase to failed, similar to below?

```go
sc.setOperationPhase(common.OperationFailed, fmt.Sprintf("Namespace auto creation failed: %s", err))
```

I suspect it's the former, just want to make sure 😄

pkg/sync/sync_context.go
```diff
@@ -810,13 +811,27 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
 					tasks = append(tasks, nsTask)
```
leoluz, line 811: I am not sure if this is still correct with the new `namespaceModifier` logic. It seems that it is appending the `nsTask` without potential labels/annotations. Shouldn't we always invoke the `sc.namespaceModifier()` in case of namespace creation?

pkg/sync/sync_context.go
```diff
@@ -810,13 +811,27 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
...
 				} else {
 					sc.log.WithValues("namespace", sc.namespace).Info("Namespace already exists")
```
leoluz, line 813: It seems that this log entry is a bit misleading. Maybe the message needs to be rephrased or moved elsewhere?

pkg/sync/sync_context.go
```diff
@@ -810,13 +811,27 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
...
+							tasks = append(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: liveObjCopy, liveObj: liveObj})
+						}
 					}
 				}
 			} else if apierr.IsNotFound(err) {
-				tasks = append(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj, liveObj: nil})
+				modified, err := sc.namespaceModifier(unstructuredObj)
+				if err != nil {
+					task := &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj}
+					sc.setResourceResult(task, common.ResultCodeSyncFailed, common.OperationError, fmt.Sprintf("Namespace auto creation failed: %s", err))
```
leoluz, line 830: Better add more context to the error message. Suggestion:
```golang
fmt.Sprintf("Namespace auto creation failed: namespace modifier error: %s", err))
```

pkg/sync/sync_context.go
leoluz: correct

pkg/sync/sync_context.go
```diff
@@ -829,6 +827,24 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
...
+	} else if modified {
+		tasks = append(tasks, preTask)
+	}
+
+	return tasks
+}
+
+func (sc *syncContext) appendFailedNsTask(tasks syncTasks, unstructuredObj *unstructured.Unstructured, err error) syncTasks {
+	task := &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj}
+	sc.setResourceResult(task, common.ResultCodeSyncFailed, common.OperationError, fmt.Sprintf("Namespace auto creation failed: error retrieving namespace from k8s: %s", err))
```
leoluz, line 843: As `appendFailedNsTask` can be called in different places this error message can be misleading. The error message needs to be added to the `err` previously.

pkg/sync/sync_context.go
```diff
@@ -829,6 +827,24 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
 	return tasks
 }
 
+func (sc *syncContext) appendNsTask(tasks syncTasks, preTask *syncTask, unstructuredObj *unstructured.Unstructured) syncTasks {
+	modified, err := sc.namespaceModifier(unstructuredObj)
+	if err != nil {
+		tasks = sc.appendFailedNsTask(tasks, unstructuredObj, err)
```
leoluz, line 833: You can add more context to the error message with:
```golang
sc.appendFailedNsTask(tasks, unstructuredObj, fmt.Errorf("namespaceModifier error: %s", err))
```

pkg/sync/sync_context.go
```diff
@@ -807,20 +808,17 @@ func (sc *syncContext) autoCreateNamespace(tasks syncTasks) syncTasks {
...
 					}
 				}
 			} else if apierr.IsNotFound(err) {
-				tasks = append(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj, liveObj: nil})
+				tasks = sc.appendNsTask(tasks, &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj, liveObj: nil}, unstructuredObj)
 			} else {
-				task := &syncTask{phase: common.SyncPhasePreSync, targetObj: unstructuredObj}
-				sc.setResourceResult(task, common.ResultCodeSyncFailed, common.OperationError, fmt.Sprintf("Namespace auto creation failed: %s", err))
-				tasks = append(tasks, task)
+				tasks = sc.appendFailedNsTask(tasks, unstructuredObj, err)
```
blakepettersson, line 821: @leoluz should I add a clarifying error message in this task as well?

pkg/sync/sync_context.go
blakepettersson: Ah right, that should go in the line under the `liveObj` nil check?

leoluz: correct!

leoluz (approved):
LGTM
Thank you for the hard work!

leoluz:
This PR introduces a breaking change in current Argo CD master. To avoid possible build issues this should be merged together with https://github.com/argoproj/argo-cd/pull/10672

PR #467, opened 2022-09-28
fix: calculate SSA diffs with smd.merge.Updater

leoluz:
This PR introduces a different strategy to calculate diffs while using server-side apply.
The new approach will delegate all diff calculation logic to the `merge.Updater.Apply` function in the `structured-merge-diff` k8s library. 

The previous implementation was naive and was missing many edge cases already addressed by `merge.Updater.Apply`.

However, `merge.Updater.Apply` has a hard to solve dependencies requirements.  In order to provide the necessary dependencies, this PR borrows code from the k8s api-server which is properly documented in the `pkg.diff.internal.fieldmanager` package (see `doc.go`).

Fix https://github.com/argoproj/argo-cd/issues/10627


pkg/diff/internal/fieldmanager/wrapper.go
```diff
@@ -0,0 +1,13 @@
+package fieldmanager
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/managedfields"
+	"sigs.k8s.io/structured-merge-diff/v4/merge"
+)
+
```
ashutosh16, line 9: add a comment here similar to the one you described regarding this function in doc.go?

zachaller (approved):

crenshaw-dev (approved):
lgtm! Gonna ask @alexmt for a review, mostly for awareness.

crenshaw-dev:
Leo spoke with Alex about this change at ArgoCon, so I'm comfortable merging it.

PR #469, opened 2022-10-03
feat: ignore endpointslice modified events, and add tests

fungiboletus:
This is related to [a comment stating that endpointslice should be ignored](https://github.com/argoproj/argo-cd/issues/9819#issuecomment-1171045786).


sidewinder12s:
This would be useful, we have a lot of DaemonSets with services that are causing constant reconciliations due to constant cluster scaling.

aaemula:
This change is awesome.
So we don't have to manually configure the `resource.exclusions`.

aaemula:
@crenshaw-dev @alexmt @jessesuen PTAL

diranged:
What needs to be done to get this over the hump? This has a serious load impact for us (https://github.com/argoproj/argo-cd/issues/15594#issuecomment-1936709854).

fungiboletus:
Is there a scenario where you don't want EndpointSlice to be ignored? I saw the code was ready to add more ignored ressources so I just added it there.

I think it's fine to have default ignored ressources in `resource.exclusions` instead. I wouldn't mind if someone else feels like implementing it :D

agaudreault:
@fungiboletus we do have them in the UI because they contain useful networking information to identify and debug some network problems. 

I'll create an issue for the 3.0 version to change the default settings :)

PR #478, opened 2022-11-03
feat:add qps and burst argument

alswl:
When using agent as benchmark tool for gitops-engine library, we must set the QPS and burst for client-go.

This PR allows the user to set up these two arguements.


alswl:
@zachaller

Hi, could you review this simple pull request?

alswl:
Friendly ping with @alexmt .

PR #493, opened 2022-12-28
feat: Expose validation webhook warnings in Argo CD UI

Sergey-Kizimov:
A warning handler has been added to this PR to display webhook validation warnings in the Argo CD user interface.
This PR is related to [Issue](https://github.com/argoproj/argo-cd/issues/9256) posted before.
Warnings in the ArgoCD UI will look like this:

Screen Shot 2022-12-28 at 2 43 40 PM
Screen Shot 2022-12-28 at 2 44 01 PM


Sergey-Kizimov:
One hack that I see here is that we can publish warning messages in the `operationState.message` field without separating by resources, in this case it isn't necessary to care which resource caused the warning because all warnings will be merged

crenshaw-dev:
I haven't dived into the code, but it looks to me like there's a `tasks` list where the result of each task is stored. Maybe that's the appropriate place to handle/persist the webhook warning response?

Issue #501, opened 2023-01-30
Respect `controller` flag in owner references


mkjpryor:
# Summary

Currently, Argo CD only considers resources with no owner references as top-level resources, and hence refuses to delete objects that have _any_ owner references.

I would like to have a configuration and/or sync option to say that only owner references with `controller: true` should be respected when determining whether a resource is a top-level resource and can be deleted by Argo.

# Motivation

I would like to use Argo CD to manage Cluster API clusters, but Cluster API adds owner references with `controller: false` to all resources that are part of a cluster but not directly managed by Cluster API (in particular, `MachineDeployment`s). This means that Argo CD refuses to delete those resources even though it created them and there are no other controllers managing them.

I believe that that this is the correct behaviour from Cluster API, as those resources do indeed have a dependency on the `Cluster` object but Cluster API is not responsible for managing their lifecycle (other than cascade deletion).

Currently, this is a massive blocker for anyone who wants to use Argo CD to manage Cluster API clusters.

# Proposal

I would like to see an additional sync option, `RespectControllerReferencesOnly=true`, that can be set on an application to enable this behaviour.

mkjpryor:
See also https://github.com/argoproj/argo-cd/issues/12210

PR #503, opened 2023-01-30
WIP: Only respect controller refs for resources

mkjpryor:
This code implements the semantics raised in #501, mainly so I could see what the effects would be. I am actually pretty happy that the result suits my use case with Cluster API quite happily.

However I am well aware that this code will break a lot of tests (mostly because none of the mock objects have `controller: true` on their owner references even when the real objects do, e.g. deployment/rs/pod) and that, because it changes the delete semantics it will never be accepted as-is.

What I would really like to have is an additional sync option that controls this behaviour, and would appreciate any advice on how to achieve a similar effect at a place in the code where I am able to use sync options.


mkjpryor:
What I am mainly concerned about is allowing Argo (i.e. gitops-engine) to delete resources that have owner references as long as none of the references have `controller: true`. However my understanding of the code is that gitops-engine identifies "top-level" resources (i.e. those without any owner references, as currently implemented) as it deploys the resource, and will only act on those resources at delete time but without checking the owner references again.

So I am struggling to see how I can get the behaviour I want at a point in the code where the sync options would be available.

mkjpryor:
@jannfis @spjmurray @jaideepr97

Just flagging that I believe this patch addresses https://github.com/argoproj/argo-cd/issues/11972 and https://github.com/argoproj/argo-cd/issues/12210.

I have been running a custom build of Argo CD that includes this patch and can confirm that it completely fixes my issue. Build is here: https://github.com/stackhpc/argo-cd/pkgs/container/argocd/67404745?tag=v2.6.0-stackhpc.2

Like I said on the issue, this is similar but subtly different to @spjmurray's solution. In short, I don't believe we should ever delete anything with a reference that has `controller: true` (@spjmurray this actually turns out to be critical for the Cluster API use case, because CAPI propagates annotations from `MachineDeployment`s to the `MachineSet`s that they create, resulting in the owner reference being ignored for the `MachineSet` as well unless controller references are always respected).

I'm happy to do whatever tidying up is required to get this merged.

javanthropus:
@mkjpryor, how does this handle node rollouts?  We produce new AWSMachineTemplates, for instance, and when we do we no longer produce the old ones.  Our MachineDeployments are then updated to reference the new templates.  This triggers the creation of new MachineSets after which the replicas are scaled over from the old MachineSets to the new ones.

The problem is that this change would immediately clobber the old AWSMachineTemplates, making the old MachineSets invalid.  While scaling them down wouldn't be an issue, scaling them up would be.  Unfortunately, scale up during a node rollout can happen and does so by scaling up the *old* MachineSet.  That will fail to spawn new nodes because of the missing template.

Have you run into this problem, and if so what do you do about it?

spjmurray:
The key problem this is solving is if it was generated by a helm template, and is no longer generated by the helm template, it gets deleted, period, do as I command!  At present it's impossible to remove a machine deployment without some 3rd party automation that goes around deleting them manually due to owner references that get added by CAPI.  I, for one, _hate_ this code because it's a) a nasty hack b) flaky as hell (for reasons I shall not go into).

I'm guessing if you want old machine templates to live on then you keep them alive in the helm chart, once your "upgrade" has been completed, then remove them.  That would be the simple solution.  Not perfect however as you'd either need to keep all templates alive forever (which happens implicitly at present) which would probably bloat your charts, and have some tooling to remove stale ones once the referencing machine deployments get killed.

Hard problem however you look at it 😸

mkjpryor:
@javanthropus 

You are right - that is a major issue. In fact, it is so major that we never delete machine templates (`OpenStackMachineTemaplate`s in our case) by using the `helm.sh/resource-policy: keep` annotation on them.

See https://github.com/stackhpc/capi-helm-charts/blob/main/charts/openstack-cluster/templates/node-group/openstack-machine-template.yaml#L61.
`

mkjpryor:
@spjmurray 

When you say

> I, for one, hate this code because it's a) a nasty hack b) flaky as hell (for reasons I shall not go into).

which specific piece of the puzzle are you referring to here?

spjmurray:
@mkjpryor 

This bit... We essentially have to:

* Derive what MDs we expect to be generated by the chart, which in itself is a pain as we have to know how the chart generates resource names
* Find all the KCTs referenced by those MDs
* Final all the OSMTs referenced by those MDs
* Delete all the resources that aren't part of those sets

I could probably do it more intelligently by adding an annotation to all the resources so we tie them to the helm inputs, rather than following the links, but it's still sub optimal when Argo should be capable of deleting them itself.

spjmurray:
@javanthropus I did try to open a discussion with CAPI directly here https://github.com/kubernetes-sigs/cluster-api/issues/7913 it's definitely on the radar, but will take time.

javanthropus:
@spjmurray, thanks for that pointer.  I piled onto that issue, and if the maintainers agree to my suggestion, I should be able to submit a PR.

dntosas:
@mkjpryor do you have any news from maintainers for this?

mkjpryor:
Not at present

dprotaso:
@crenshaw-dev can you take a look at this PR?

todaywasawesome:
Subscribed...

dprotaso:
I added this issue/PR to the ArgoCD monthly community meeting. You can see the notes here: https://docs.google.com/document/d/1ttgw98MO45Dq7ZUHpIiOIEfbyeitKHNfMjbY5dLLMKQ/edit?tab=t.0#heading=h.qamcy9ybyb10 

I won't be able to attend (on vacation). If someone  (maybe @mkjpryor?) could attend Wed Dec 4th 10am PST to surface this PR in a discussion that would be great.

dprotaso:
Hey @mkjpryor I followed up with some of the maintainers in the original slack thread here:

https://cloud-native.slack.com/archives/C01TSERG0KZ/p1750115707909679?thread_ts=1675116027.741979&cid=C01TSERG0KZ

It seems like the approach in the PR is good - are you able to resolve the merge conflicts?

dmosesson:
At what point does someone else have the ability to resolve these merge conflicts?

mkjpryor:
I'm afraid I have moved on from the company whose organisation the fork is in, and am not working on the product in which this was an issue any more.

As such I'm unable to commit to a time when I might get to rebasing this. If somebody else has time then I'm happy for them to go ahead and do it in another PR.

dprotaso:
rebase is here: https://github.com/argoproj/gitops-engine/pull/771

crenshaw-dev:
Closing in favor of the rebase PR. Thanks @mkjpryor for the original PR and @dprotaso for picking it up!

PR #514, opened 2023-03-19
feat: Add sync-dependencies

alexec:
Depended on by [argo-cd#3517](https://github.com/argoproj/argo-cd/issues/3517)

This PR introduces a new feature: sync dependencies.

Sync dependencies give a new way to order sync operations that is influenced by sync waves and sync hooks (all of which I wrote, so I know it extremely well). 

Like sync waves, a sync operation will only progress when all the dependents have completed. 

Dependencies between objects is specified by the new `argocd.argoproj.io/sync-dependencies` annotation. 

```yaml
# The dependencies are specified as a comma-separated list of objects.
# The format of each object is <group>/<kind>/<namespace>/<name>
# The group and kind are optional. If not specified, they'll match any group or kind.
# The namespace is optional. If not specified, the namespace is assumed to be the same as the namespace of the object.
# The name is required.
argocd.argoproj.io/sync-dependencies: a,b,c
```

Reasons to do this:

* The 3rd most popular enhancement request for Argo CD
* Explanation can be found in [argo-cd#3517](https://github.com/argoproj/argo-cd/issues/3517).
* Implementation is contained within the GitOps engine.

Reasons not to do this:

* Impacts core GitOps engine code (must be mitigated by exceptional testing).
* Another gun to shoot yourself in the foot with.

Open questions:

* Sync waves, hooks, and dependencies have a certain level of opacity to them. By which I mean, that it is not always clear what they're doing unless you're familiar with the app. I wonder if we should introduce a new "waiting" state for resources?
* Should dependencies take precedence over waves? I think this is yes.
* How should dependencies be described? I've used annotation, is that right? Is that the right structure? 
* Are the rules that match tasks to dependencies?

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -480,7 +480,7 @@ func (sc *syncContext) Sync() {
 
 	sc.log.WithValues("tasks", tasks).V(1).Info("Filtering out non-pending tasks")
 	// remove tasks that are completed, we can assume that there are no running tasks
-	tasks = tasks.Filter(func(t *syncTask) bool { return t.pending() })
+	completedTasks, tasks := tasks.Split(func(t *syncTask) bool { return t.completed() })
 
 	if sc.applyOutOfSyncOnly {
 		tasks = sc.filterOutOfSyncTasks(tasks)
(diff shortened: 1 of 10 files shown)


pkg/sync/sync_context.go
alexec, line 1: In `syncTasks` we need to check for cyclic-dependencies.

pkg/sync/sync_context.go
```diff
@@ -500,13 +500,19 @@ func (sc *syncContext) Sync() {
...
-	sc.log.WithValues("phase", phase, "wave", wave, "tasks", tasks, "syncFailTasks", syncFailTasks).V(1).Info("Filtering tasks in correct phase and wave")
-	tasks = tasks.Filter(func(t *syncTask) bool { return t.phase == phase && t.wave() == wave })
+	tasks, blockedTasks := tasks.Split(func(t *syncTask) bool {
+		for _, dep := range t.dependencies() {
+			if !completedTasks.Any(func(t *syncTask) bool { return dep.match(t.obj()) }) {
+				sc.log.WithValues("task", t, "dependency", dep).V(1).Info("Task has unmet dependency")
+				return false
+			}
+		}
+		return t.phase == phase && wave == t.wave()
```
alexec, line 513: This does not take into account `isHook`.

pkg/sync/sync_context.go
```diff
@@ -529,12 +535,12 @@ func (sc *syncContext) Sync() {
...
 		sc.setOperationFailed(syncFailTasks, syncFailedTasks, "one or more objects failed to apply")
 	case successful:
-		if remainingTasks.Len() == 0 {
+		if blockedTasks.Len() == 0 {
 			// delete all completed hooks which have appropriate delete policy
 			sc.deleteHooks(hooksPendingDeletionSuccessful)
 			sc.setOperationPhase(common.OperationSucceeded, "successfully synced (all tasks run)")
 		} else {
-			sc.setRunningPhase(remainingTasks, false)
+			sc.setRunningPhase(tasks, false)
```
alexec, line 543: Looks like a 2 year old bug.

pkg/sync/sync_context_test.go
```diff
@@ -1133,6 +1132,7 @@ func TestSyncFailureHookWithFailedSync(t *testing.T) {
 }
 
 func TestBeforeHookCreation(t *testing.T) {
+	t.SkipNow()
```
alexec, line 1135: revert

pkg/sync/sync_context.go
```diff
@@ -772,6 +786,8 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
 
 	tasks.Sort()
 
+	sc.log.WithValues("tasks", tasks).V(1).Info("tasks after sorting")
+
```
alexec, line 790:
```suggestion
```

reggie-k:
I wonder whether we can hit a scenario of a dependent resource that has an auto-generated and thus unknown ahead of time name?

crenshaw-dev:
@or-shachar see discussion here: https://github.com/argoproj/argo-cd/issues/7437

I don't think there's been sufficient work done on the proposal side to show that the feature is necessary/justified or describing the details of how it will function.

Gonna close this for now, since I don't think Alex is actively pursuing the feature.

Issue #525, opened 2023-05-30
Extend sync-options annotation with `Force=true` 


n9:
Solution for https://github.com/argoproj/argo-cd/issues/5882

Use case: e.g. to replace certain pod

(I will try to implement it.)

morey-tech:
This issue might be a duplicate of https://github.com/argoproj/gitops-engine/issues/414, you could close this and link your PR to the older one

Issue #414, opened 2022-05-30
Add annotation `argocd.argoproj.io/sync-options: Force=true`


jiwonaid:
## Summary
This issue is about the need for the new annotation `argocd.argoproj.io/sync-options: Force=true`, which is needed for the use case like the job resources that should run every time when syncing.

## Goals & Proposal

Could we implement the new annotation feature on this repo?

Based on `kubectl replace --help`, `--force` is needed to delete and then re-create the resources.
```bash
~  kubectl replace --help
  
  # Force replace, delete and then re-create the resource
  kubectl replace --force -f ./pod.json
  
```

Currently, there is nothing like `argocd.argoproj.io/sync-options: Force=true`.
It seems like `argocd.argoproj.io/sync-options: Replace=true` runs on this library.
https://github.com/argoproj/gitops-engine/blob/b855894da06d5bcb2f26ebd3aebd92f1591dbe0a/pkg/sync/sync_context.go#L176-L180

## Releated Issues
There are some related issues at argo-cd repo. Please take a look.
https://github.com/argoproj/argo-cd/issues/5882
https://github.com/argoproj/argo-cd/issues/7459
https://github.com/argoproj/argo-cd/issues/9163

Thank you for reading.

jelical:
Any progress with this enhancement?

jiwonaid:
Hi @jelical,
I am not sure argocd's current status, and not working on this.

Anyone who would like to push this forward can take this.

Thank you.

sidharthramesh:
Is there a way in ArgoCD to do this currently?

thesuperzapper:
@renshaw-dev @gdsoumya @alexec I think this is an absolutely critical feature, so would appreciate your feedback.

Currently, ArgoCD doesn't have a way to use `kubectl apply --force` or `kubectl replace --force`, and this is required to update some common types of Kubernetes resources, like Jobs and CronJobs and even bare Pods.

The solution posed here is very elegant, as it allows users to specifically annotate `argocd.argoproj.io/sync-options: Replace=true,Force=true` on resources that they know can't be updated without being deleted first.

thesuperzapper:
@crenshaw-dev @gdsoumya @alexec I think this is an absolutely critical feature, so would appreciate your feedback.

Currently, ArgoCD doesn't have a way to use `kubectl apply --force` or `kubectl replace --force`, and this is required to update some common types of Kubernetes resources, like Jobs and CronJobs and even bare Pods.

[...]

PR #526, opened 2023-05-30
feat: Extend sync-options annotation with Force=true (#525)

n9:
Implementation for #525, #414, https://github.com/argoproj/argo-cd/issues/5882

### DIFF
--- a/pkg/sync/common/types.go
+++ b/pkg/sync/common/types.go
@@ -27,6 +27,8 @@ const (
 	SyncOptionPruneLast = "PruneLast=true"
 	// Sync option that enables use of replace or create command instead of apply
 	SyncOptionReplace = "Replace=true"
+	// Sync option to use with replace that enables resource recreation (e.g. pod)
+	SyncOptionForce = "Force=true"
 	// Sync option that enables use of --server-side flag instead of client-side
 	SyncOptionServerSideApply = "ServerSideApply=true"
 	// Sync option that disables resource deletion
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -903,7 +903,7 @@ func (sc *syncContext) ensureCRDReady(name string) error {
 	})
 }
 
-func (sc *syncContext) applyObject(t *syncTask, dryRun, force, validate bool) (common.ResultCode, string) {
+func (sc *syncContext) applyObject(t *syncTask, dryRun, validate bool) (common.ResultCode, string) {
 	dryRunStrategy := cmdutil.DryRunNone
 	if dryRun {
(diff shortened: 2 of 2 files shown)


crenshaw-dev (commented):
Looks good! Is there any existing unit testing for Replace which we could replicate for Force?

n9:
@crenshaw-dev I have found: https://github.com/argoproj/gitops-engine/blob/c0ffe8428a23011ae9f6d0d3d01f9bf6200a4339/pkg/sync/sync_context_test.go#L755

But it seems to me that it just tests that the correct method is used.

How the "force" flag is tested?

sidharthramesh:
Any updates on this?

crenshaw-dev:
@n9 good question... I'm not sure how `resourceOps.GetLastResourceCommand(kube.GetResourceKey(tc.target)` works, but basically I'm looking for a way to verify that, when the Force=true option is specified, the kubectl action is actually performed with `--force` enabled.

crenshaw-dev:
@n9 I believe the mock `ApplyResource` function could be modified to store that flag information. 

```go
func (r *MockResourceOps) ApplyResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string) (string, error) {
	r.SetLastValidate(validate)
	r.SetLastServerSideApply(serverSideApply)
	r.SetLastServerSideApplyManager(manager)
	r.SetLastResourceCommand(kube.GetResourceKey(obj), "apply")
	command, ok := r.Commands[obj.GetName()]
	if !ok {
		return "", nil
	}

	return command.Output, command.Err
}
```

mcoklica1:
Hey @n9 @crenshaw-dev, any updates on testing this and proceeding forward? IIUC you want to see if force flag is visible via ArgoCD logs when sync happens (application-controller pod)?

n9:
Unfortunately, I have to work on other things. Feel free to complete this PR.

kkk777-7:
@crenshaw-dev Implemented unit test. Could you please take a look at [this PR](https://github.com/argoproj/gitops-engine/pull/560) when you have some time? Thank you.

ChristianCiach:
This seems to be obsolete now that https://github.com/argoproj/gitops-engine/pull/560 has been merged, I think.

ktzsolt:
Hi! 

First of all, thank you for the change! 

May I ask how (and when) will this be integrated into Argo CD? All I see for gitops-engine is that the latest release is 2 years old here https://pkg.go.dev/github.com/argoproj/gitops-engine?tab=versions
So I don't quite get the release management of this component.

PR #532, opened 2023-07-06
feat: auto respect rbac for discovery/sync

gdsoumya:
This PR implements this [proposal](https://github.com/argoproj/argo-cd/blob/master/docs/proposals/respect-rbac-for-resource-exclusions.md). PR enables gitops engine to automatically stop watching for resources that it does not have permission to access without having to manually set resource exclusions.

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -11,9 +11,9 @@ import (
 
 	"github.com/go-logr/logr"
 	"golang.org/x/sync/semaphore"
-	authorizationv1 "k8s.io/api/authorization/v1"
 	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -23,8 +23,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/dynamic"
-	"k8s.io/client-go/kubernetes"
-	authType1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/pager"
@@ -467,10 +465,6 @@ func (c *clusterCache) startMissingWatches() error {
 	if err != nil {
 		return err
 	}
-	clientset, err := kubernetes.NewForConfig(c.config)
-	if err != nil {
-		return err
-	}
 	namespacedResources := make(map[schema.GroupKind]bool)
 	for i := range apis {
 		api := apis[i]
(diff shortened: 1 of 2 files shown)


pkg/cache/cluster.go
```diff
@@ -702,6 +741,51 @@ func (c *clusterCache) processApi(client dynamic.Interface, api kube.APIResource
 	return nil
 }
 
+// checkPermission runs a self subject access review to check if the controller has permissions to list the resource
+func (c *clusterCache) checkPermission(ctx context.Context, reviewInterface authType1.SelfSubjectAccessReviewInterface, api kube.APIResourceInfo) (keep bool, err error) {
+	sar := &authorizationv1.SelfSubjectAccessReview{
+		Spec: authorizationv1.SelfSubjectAccessReviewSpec{
+			ResourceAttributes: &authorizationv1.ResourceAttributes{
+				Namespace: "*",
+				Verb:      "list", // uses list verb to check for permissions
```
jannfis, line 750: Is `list` sufficient? Or would we also need to check for `get` and `watch`?

pkg/cache/cluster.go
```diff
@@ -702,6 +741,51 @@ func (c *clusterCache) processApi(client dynamic.Interface, api kube.APIResource
...
+				Resource:  api.GroupVersionResource.Resource,
+			},
+		},
+	}
+
+	switch {
+	// if manage whole cluster or resource is cluster level and cluster resources enabled
+	case len(c.namespaces) == 0 || !api.Meta.Namespaced && c.clusterResources:
```
jannfis, line 758:
```suggestion
	case len(c.namespaces) == 0 || (!api.Meta.Namespaced && c.clusterResources):
```

I know that this is implicit by operator precedence, but imho explicit braces make it much easier to understand the condition

jannfis (commented):
Initial round of review, and I have some comments and questions.

PTAL.

pkg/cache/cluster.go
gdsoumya: sure will update, this condition was taken from an existing function `processApi` so will update it there too.

pkg/cache/cluster.go
```diff
@@ -702,6 +741,51 @@ func (c *clusterCache) processApi(client dynamic.Interface, api kube.APIResource
...
+		resp, err := reviewInterface.Create(ctx, sar, metav1.CreateOptions{})
+		if err != nil {
+			return true, err
+		}
```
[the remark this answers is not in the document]
gdsoumya: actually it's an inconclusive response, we could say false too but as the request errored either of those values can be possible I chose true because instead of denying without confirmation the function responds with confirmation when error occurs. Though currently, it doesn't matter as the error will always take precedence.

pkg/cache/cluster.go
gdsoumya: Also, the error response after which this access review is made is a list call thus it makes sense to use list verb to confirm the access error.

pkg/cache/cluster.go
```diff
@@ -748,6 +832,10 @@ func (c *clusterCache) sync() error {
 	if err != nil {
 		return err
 	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
gdsoumya: Tested it with dynamic client and for some reason the request to create the access review resource failed with an unauthorized error. I wasn't able to determine why it was happening but my best guess is the dynamic client needs a namespace for creating the resource, but access review is a cluster scope resource so when I pass "" as the namespace it fails the request.

pkg/cache/cluster.go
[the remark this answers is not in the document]
gdsoumya: updated to false

pkg/cache/cluster.go
jannfis: Makes sense.

alexmt (approved):
LGTM!

PR #535, opened 2023-07-26
Drop dependency on k8s.io/kubernetes

### LINKED ISSUE #56: Drop k8s.io/kubernetes dependency

cehoffman:
This fixes #56 and helps on the path to addressing https://github.com/argoproj/argo-cd/issues/14727

The go.mod and go.sum changes are from running `go mod tidy`.


pkg/utils/kube/scheme/scheme.go
```diff
@@ -1,29 +1,106 @@
...
+	storagev1 "k8s.io/api/storage/v1"
+	storagev1alpha1 "k8s.io/api/storage/v1alpha1"
+	storagev1beta1 "k8s.io/api/storage/v1beta1"
 )
 
-var Scheme = legacyscheme.Scheme
+var Scheme = runtime.NewScheme()
+
+func init() {
+	runtimeutil.Must(admissionv1.AddToScheme(Scheme))
```
alexmt, line 59: I did not know it was possible to get defaults from `k8s.io/api` package 🤯.

alexmt:
Thanks a lot for PR @cehoffman ! Removing conversion will affect PR but probably we can do something about it. E.g. cache converted version as well. Let me please test and see if significantly affects performance

pkg/utils/kube/scheme/scheme.go
```diff
@@ -1,29 +1,106 @@
 package scheme
 
 import (
-	"k8s.io/kubernetes/pkg/api/legacyscheme"
+	"k8s.io/apimachinery/pkg/runtime"
```
blakepettersson, line 4: Would it be possible to instead use the Scheme found in `k8s.io/client-go/kubernetes/scheme` (there's probably a good reason it cannot, but would like to understand why)?

cehoffman: I think that would actually be preferred. Thanks for pointing to it. It is equivalent.

cehoffman: done

majolo:
@cehoffman @alexmt is this PR still active? We would benefit from it also - happy to help where I can.

mikejoh:
> @cehoffman @alexmt is this PR still active? We would benefit from it also - happy to help where I can.

Same here!

PR #537, opened 2023-08-17
fix: apply ssa namespace twice

blakepettersson:
In argoproj/argo-cd#13965, upgrading to client-go 0.26 uncovered an issue with existing namespaces not preserving labels and annotations when converting an existing namespace to use `managedNamespaceMetadata`.

From 0.26 up, in order to preserve the existing behaviour what we need to do is to first apply the existing namespace with SSA and its preexisting labels and annotations. This also needs to be applied with another manager than the default one.

After that initial apply, we proceed as normal with a ServerSideApply. This is a one-time thing for any existing namespace; we check if we need to perform this action by detecting if there is a preexisting SSA annotation on the namespace.


pkg/sync/sync_context.go
```diff
@@ -934,6 +934,20 @@ func (sc *syncContext) applyObject(t *syncTask, dryRun, force, validate bool) (c
...
+		liveObjIsNotSSA := t.liveObj != nil && !resourceutil.HasAnnotationOption(t.liveObj, common.AnnotationSyncOptions, common.SyncOptionServerSideApply)
+		if t.targetObj.GetKind() == kube.NamespaceKind && t.liveObj != nil && t.liveObj.GetKind() == kube.NamespaceKind && targetObjIsSSA && liveObjIsNotSSA {
+			// If there is a pre-existing live namespace, and if we also want to use managedNamespaceMetadata on an existing namespace,
+			// we need to do `kubectl apply` with the namespace as-is, using a different manager than the default one.
+			// The reason we do this is so that we can preserve existing labels and annotations on the namespace.
+			nsSpec := &v1.Namespace{TypeMeta: metav1.TypeMeta{APIVersion: "v1", Kind: kube.NamespaceKind}, ObjectMeta: metav1.ObjectMeta{Name: t.liveObj.GetName(), Annotations: t.liveObj.GetAnnotations(), Labels: t.liveObj.GetLabels(), Namespace: t.liveObj.GetNamespace()}}
+			liveCopy, err := kube.ToUnstructured(nsSpec)
+
+			message, err = sc.resourceOps.ApplyResource(context.TODO(), liveCopy, dryRunStrategy, force, validate, serverSideApply, "argocd-controller-tmp")
```
blakepettersson, line 946: `argocd-controller-tmp` is a placeholder name, but the idea is that in order to preserve existing metadata we need to apply the namespace with a separate manager before using `managedNamespaceMetadata` on said namespace.

pkg/sync/sync_context.go
```diff
@@ -934,6 +934,23 @@ func (sc *syncContext) applyObject(t *syncTask, dryRun, force, validate bool) (c
...
+		targetObjIsSSA := resourceutil.HasAnnotationOption(t.targetObj, common.AnnotationSyncOptions, common.SyncOptionServerSideApply)
+		liveObjIsNotSSA := t.liveObj != nil && !resourceutil.HasAnnotationOption(t.liveObj, common.AnnotationSyncOptions, common.SyncOptionServerSideApply)
+		if t.targetObj.GetKind() == kube.NamespaceKind && t.liveObj != nil && t.liveObj.GetKind() == kube.NamespaceKind && targetObjIsSSA && liveObjIsNotSSA {
+			// If there is a pre-existing live namespace, and if we also want to use managedNamespaceMetadata on an existing namespace,
+			// we need to do `kubectl apply` with the namespace as-is, using a different manager than the default one.
+			// The reason we do this is so that we can preserve existing labels and annotations on the namespace.
+			nsSpec := &v1.Namespace{TypeMeta: metav1.TypeMeta{APIVersion: "v1", Kind: kube.NamespaceKind}, ObjectMeta: metav1.ObjectMeta{Name: t.liveObj.GetName(), Annotations: t.liveObj.GetAnnotations(), Labels: t.liveObj.GetLabels(), Namespace: t.liveObj.GetNamespace()}}
+			liveCopy, err := kube.ToUnstructured(nsSpec)
+			if err != nil {
+				return common.ResultCodeSyncFailed, err.Error()
```
crenshaw-dev, line 946:
```suggestion
				return common.ResultCodeSyncFailed, "failed to convert namespace spec to unstructured: " + err.Error()
```

pkg/sync/sync_context.go
```diff
@@ -934,6 +934,23 @@ func (sc *syncContext) applyObject(t *syncTask, dryRun, force, validate bool) (c
...
+			}
+
+			_, err = sc.resourceOps.ApplyResource(context.TODO(), liveCopy, dryRunStrategy, force, validate, serverSideApply, "argocd-controller-tmp")
+			if err != nil {
+				return common.ResultCodeSyncFailed, err.Error()
```
crenshaw-dev, line 951:
```suggestion
				return common.ResultCodeSyncFailed, "failed to apply existing namespace with temp field manager: " + err.Error()
```

crenshaw-dev (changes requested):

PR #541, opened 2023-09-29
chore: log cluster sync error

crenshaw-dev:
Right now we only pass this up to the caller to log. That has two downsides:
1) The caller may not log it (there's at least one such case in Argo CD)
2) The caller may log it _too often_, producing duplicate log lines and making it difficult to determine exactly how many clusters are "sick" - for example, Argo CD logs this error for every single Application deployed to the sick cluster


leoluz (approved):
LGTM

PR #547, opened 2023-10-31
Revert "fix(server): use server side dry run in case if it is server side apply"

pasha-codefresh:
Reverts argoproj/gitops-engine#546


pasha-codefresh:
Снимок экрана 2023-10-31 в 18 16 54
This issue started to appear after this change

PR #554, opened 2023-11-27
chore: more logging for CRD updates

crenshaw-dev:
CRD updates are memory/CPU expensive. If CRDs are changing often, that can cause performance issues. Log changes to help users diagnose performance issues.


jannfis (approved):
LGTM

gdsoumya (approved):

ishitasequeira (approved):

PR #559, opened 2023-12-05
bug: detect and report cluster connection errors

chetan-rns:
Currently, Argo CD doesn't report connection errors on the cluster UI page unless the cache is invalidated (manually/periodically every 24 hours) or there is a sync error. During auto-sync, this error is only updated if there is a new commit to sync. This PR adds a new field `ConnectionStatus` that gets updated whenever Argo CD fails to access the cluster. The [ClusterInfoUpdater](https://github.com/argoproj/argo-cd/blob/master/controller/clusterinfoupdater.go) will periodically fetch this cluster info and update the `appstate` cache.


pkg/cache/cluster.go
```diff
@@ -615,6 +630,25 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
...
+				if err != nil {
+					if c.connectionStatus != ConnectionStatusFailed {
+						c.log.Info("unable to access cluster", "cluster", c.config.Host, "reason", err.Error())
+						c.connectionStatus = ConnectionStatusFailed
+						connectionUpdated = true
+					}
+				} else if c.connectionStatus != ConnectionStatusSuccessful {
+					c.connectionStatus = ConnectionStatusSuccessful
+					connectionUpdated = true
+				}
```
jgwest, line 643: The `clusterCache` lock likely needs to be owned before modifying `c.connectionStatus` within this method. We just need to make sure the lock is released before calling `c.Invalidate()`, as default go locks are not reentrant.

jgwest (changes requested):

pkg/cache/cluster.go
chetan-rns: Updated :+1:  I haven't used the locks for updating [this block](https://github.com/argoproj/gitops-engine/pull/559/files#diff-9c9e197d543705f08c9b1bc2dc404a55506cfc2935a988e6007d248257aadb1aR851) because I noticed that the locks were acquired in the parent function of sync().

pkg/cache/cluster.go
```diff
@@ -615,6 +630,29 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
 				if errors.IsNotFound(err) {
 					c.stopWatching(api.GroupKind, ns)
 				}
+				var connectionUpdated bool
+				if err != nil {
```
jgwest, line 634: Hey Chetan, me again, a few more thoughts while thinking about the ramifications of the cluster invalidation `c.Invalidate()` and `c.EnsureSynced()` calls. (Feel free to poke holes in my logic!)

- The watch function (defined here) is called for each watched resource, so if there are 100 resources, it will be called once for each, to establish the watch.
- What about the scenario where only 1 watch (or a small number are failing), for example because the ServiceAccount does not have the correct permissions to watch the resource?
	- If one watch fails 1 out of 100, will the cluster oscillate between Successful and Failed?
	- It might look like this:
	- Watch 1: succeeds, set cluster status to Successful
	- Watch 2: succeeds, set cluster status to Successful
	- Watch 3: succeeds, set cluster status to Successful
	- (...)
	- Watch 80: succeeds, set cluster status to Failed
	- Perhaps the ideal behaviour would be to report success/failure only after all were established. (Easier said than done, I'm sure)
	
- Another question: since it appears we are invalidating the cluster cache on any failure:
	- Will this cause cache invalidation and re-sync to occur much more often, if the network connection between Argo CD and the target (watched) cluster happens to be unstable? (for example, imagine a case where 1% of connections fail. Would this magnify to invalidating and forcing a resync of all the resources, if only 1% of connections were failing?).

jgwest (changes requested):

pkg/cache/cluster.go
chetan-rns: Hi Jonathan, your concerns are valid. Since watches could fail for any reason it is difficult to determine the cluster status based on the errors returned during watch. These errors could be transient making it complicated to keep track of their states. Also, the error message could be generic to determine why exactly the watch failed.

To fix this issue, I have included a goroutine that periodically checks for watch errors, pings the remote cluster(to get the version), and updates the status. By following this approach we don't rely completely on the watches and prevent the status from oscillating in the case of transient watch failures. We use the watch status to prevent the number of pings to the remote clusters by pinging the cluster only when the watches fail. I've also added checks to ensure we are invalidating the cluster cache only when the status changes which shouldn't happen frequently. 

Let me know what you think. The periodic goroutine approach does handle some of the edge cases that may arise due to the dynamic nature of the watches. But also I'm open to hearing other approaches to solve this problem.

jgwest (changes requested):
I would love to rewrite the cluster cache logic entirely, and replace the locks with go channels (which are much easier to reason against, when done right!), but that's probably not happening any time soon. :smile:

pkg/cache/cluster.go
```diff
@@ -1192,3 +1230,89 @@ func (c *clusterCache) GetClusterInfo() ClusterInfo {
...
+			return
+		}
+	}
+
+}
+
+func (c *clusterCache) updateConnectionStatus(status ConnectionStatus) {
+	if c.connectionStatus == status {
+		return
+	}
```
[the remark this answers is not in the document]
chetan-rns: Updated with the lock.

pkg/cache/cluster.go
```diff
@@ -132,6 +134,8 @@ type ClusterCache interface {
 	OnResourceUpdated(handler OnResourceUpdatedHandler) Unsubscribe
 	// OnEvent register event handler that is executed every time when new K8S event received
 	OnEvent(handler OnEventHandler) Unsubscribe
+	// UpdateClusterConnectionStatus checks the watch errors periodically and updates the cluster connection status.
+	UpdateClusterConnectionStatus(ctx context.Context)
```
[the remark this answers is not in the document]
chetan-rns: Makes sense. Renamed the function to  `StartClusterConnectionStatusMonitoring`.

pkg/cache/cluster.go
```diff
@@ -1192,3 +1230,89 @@ func (c *clusterCache) GetClusterInfo() ClusterInfo {
...
+					c.updateConnectionStatus(ConnectionStatusSuccessful)
+				}
+			}
+		case <-ctx.Done():
+			ticker.Stop()
+			return
+		}
+	}
+
+}
```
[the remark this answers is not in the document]
chetan-rns: I've refactored the function with locks to read the c.connectionStatus. Managed to avoid locking around `GetServerVersion`. Hopefully, this should fix the issue.

pkg/cache/cluster.go
```diff
@@ -1192,3 +1230,89 @@ func (c *clusterCache) GetClusterInfo() ClusterInfo {
...
+
+// UpdateClusterConnectionStatus starts a goroutine that checks for watch failures.
+// If there are any watch errors, it will periodically ping the remote cluster
+// and update the cluster connection status.
+func (c *clusterCache) UpdateClusterConnectionStatus(ctx context.Context) {
+	go c.clusterConnectionService(ctx)
+}
+
+func (c *clusterCache) clusterConnectionService(ctx context.Context) {
+	clusterConnectionTimeout := 10 * time.Second
```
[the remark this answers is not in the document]
chetan-rns: I've used 10 seconds to poll the status of watch failures. We might have to ping the remote cluster every 10s if the watches are failing. But the function `updateConnectionStatus` will run only once when the status changes. For the subsequent polls, the condition checks ensure we don't call it often. 

These are the conditions that we check before invalidating the cache:
    1. Check if there are watch failures or if it has recovered back(no watch errors but the connection status is 'Failed')
    2. If either of the above conditions are met we ping the remote cluster to get the version.
    3. If there is no error, but the connection status is `Failed` we invalidate the cache.
    4. If there is an error, but the connection status is `Successful` we invalidate the cache.
    
I'm open to updating to a longer duration. But the tradeoff is we can't recognize the failures early. This should be okay I suppose, since Argo CD will not be primarily used for cluster monitoring.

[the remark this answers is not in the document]
jannfis: I'd recommend making the interval configurable per cache, so it can be parametrized from a consumer

pkg/cache/cluster.go
```diff
@@ -1192,3 +1230,101 @@ func (c *clusterCache) GetClusterInfo() ClusterInfo {
...
+				if c.connectionStatus == ConnectionStatusFailed {
+					watchesRecovered = true
+				}
+				c.lock.RUnlock()
+			}
+
+			if watchErrors > 0 || watchesRecovered {
+				c.log.V(1).Info("verifying cluster connection", "watches", watchErrors)
+
+				_, err := c.kubectl.GetServerVersion(c.config)
```
jannfis, line 1265: I feel evaluation of this call should be done with a certain degree of tolerance for intermittent failures, e.g. a (configurable) retry on certain errors such as a flaky network.

chetan-rns: Since we call this using a ticker, it will be evaluated every X seconds (default 10). So, it will be called again after an interval in the case of intermittent failures. Do we still need to explicitly retry?  In that case, can we reuse the existing [listRetryFunc](https://github.com/argoproj/gitops-engine/blob/5fd9f449e75774f6b2b7b1ff137c7759e075e385/pkg/cache/cluster.go#L203)?

pkg/cache/cluster.go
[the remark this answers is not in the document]
chetan-rns: Made the interval configurable :+1:

pkg/cache/cluster.go
jannfis: It depends - will a failing version call lead to cluster cache invalidation, and then be re-built in the next interval?

chetan-rns: There is a rare possibility of frequent invalidations if both the `GetServerversion()` and the watches fail/recover between successive ticker intervals. I have updated the call within a retry that checks for transient network errors using [isTransientNetworkErr()](https://github.com/argoproj/argo-cd/blob/44da2063c7e915d0fc8342cdd834e6869b2910bc/controller/cache/cache.go#L391). It should retry in the case of flaky network errors without invalidating the cache.

jannfis (approved):
LGTM, thank you @chetan-rns !

PR #561, opened 2023-12-18
Object-level prune propagation policy, propagation policy for kubectl replace

sthomson-wyn:
- add object-level prune propagation configuration
- add cascading option for kubectl replace

Fixes https://github.com/argoproj/argo-cd/issues/8501


agershman:
When do we expect this change to land?

PR #562, opened 2024-01-12
fix: address kubectl auth reconcile during server-side diff

leoluz:
fix: https://github.com/argoproj/argo-cd/issues/16840

### DIFF
--- a/pkg/utils/kube/resource_ops.go
+++ b/pkg/utils/kube/resource_ops.go
@@ -92,33 +92,33 @@ func (k *kubectlResourceOperations) runResourceCommand(ctx context.Context, obj 
 
 	var out []string
 	if obj.GetAPIVersion() == "rbac.authorization.k8s.io/v1" {
-		outReconcile, err := k.rbacReconcile(ctx, obj, manifestFile.Name(), dryRunStrategy)
+		outReconcile, err := k.rbacReconcile(ctx, obj, manifestFile.Name(), dryRunStrategy, serverSideDiff)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/utils/kube/resource_ops.go
```diff
@@ -131,6 +124,30 @@ func (k *kubectlResourceOperations) runResourceCommand(ctx context.Context, obj
...
+// `auth reconcile` will delete and recreate the resource if necessary
+func (k *kubectlResourceOperations) rbacReconcile(ctx context.Context, obj *unstructured.Unstructured, fileName string, dryRunStrategy cmdutil.DryRunStrategy) (string, error) {
+	outReconcile, err := func() (string, error) {
+		cleanup, err := k.processKubectlRun("auth")
+		if err != nil {
+			return "", err
+		}
+		defer cleanup()
+		return k.authReconcile(ctx, obj, fileName, dryRunStrategy)
+	}()
```
crenshaw-dev, line 144: Does this still need to be wrapped in a function? I'm guessing the function existed before to trigger the cleanup call, but this function should behave the same whether or not we wrap the contents in an extra function.

crenshaw-dev (changes requested):

pkg/utils/kube/resource_ops.go
leoluz: I extracted the code in an external function for clarity but didn't change it. Your suggestion makes sense. Changed.

crenshaw-dev (approved):

PR #564, opened 2024-01-19
fix(server): Dry run always in client mode just for yaml manifest validation even with server side apply

anandf:
#### Root Cause: 
When server side apply is set to true, dry run is run in server mode. Dry run in server mode will fail if the target namespace is missing. So in cases where the sync option `CreateNamespace` is set to true, the dry run fails in server mode, and then attempts to re-run in client mode as a fallback mechanism. But the server side apply is still set. This is not a supported configuration in `kubectl` when `--dry-run` is set to client, `--server-side` flag is not supported causing that to fail as well with the below error message
```
# kubectl apply -k kustomize-guestbook --dry-run=client --server-side -n guestbook -o json
error: --dry-run=client doesn't work with --server-side (did you mean --dry-run=server instead?)
``` 
Due to this behaviour, when both `CreateNamespace=true` and `ServerSideApply=true` sync options are used, then the app sync fails while trying to validate the rendered yaml manifests.

#### Solution
If it is a dry run, always run the dry run in client mode `--dry-run=client` even if the server side apply option is set `--server-side`. Dry runs are used for validating if the rendered manifests satisfy the yaml correctness. For the actual sync, honour the `--server-side` apply flag.

Fixes 
https://github.com/argoproj/argo-cd/issues/16829 
https://github.com/argoproj/argo-cd/issues/13874
https://github.com/argoproj/argo-cd/issues/16840

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -921,39 +921,29 @@ func (sc *syncContext) applyObject(t *syncTask, dryRun, force, validate bool) (c
 	var err error
 	var message string
 	shouldReplace := sc.replace || resourceutil.HasAnnotationOption(t.targetObj, common.AnnotationSyncOptions, common.SyncOptionReplace)
-	applyFn := func(dryRunStrategy cmdutil.DryRunStrategy) (string, error) {
-		if !shouldReplace {
-			return sc.resourceOps.ApplyResource(context.TODO(), t.targetObj, dryRunStrategy, force, validate, serverSideApply, sc.serverSideApplyManager, false)
-		}
-		if t.liveObj == nil {
-			return sc.resourceOps.CreateResource(context.TODO(), t.targetObj, dryRunStrategy, validate)
-		}
-		// Avoid using `kubectl replace` for CRDs since 'replace' might recreate resource and so delete all CRD instances.
-		// The same thing applies for namespaces, which would delete the namespace as well as everything within it,
-		// so we want to avoid using `kubectl replace` in that case as well.
-		if kube.IsCRD(t.targetObj) || t.targetObj.GetKind() == kubeutil.NamespaceKind {
-			update := t.targetObj.DeepCopy()
-			update.SetResourceVersion(t.liveObj.GetResourceVersion())
-			_, err = sc.resourceOps.UpdateResource(context.TODO(), update, dryRunStrategy)
-			if err != nil {
-				return fmt.Sprintf("error when updating: %v", err.Error()), err
+	if shouldReplace {
+		if t.liveObj != nil {
+			// Avoid using `kubectl replace` for CRDs since 'replace' might recreate resource and so delete all CRD instances.
+			// The same thing applies for namespaces, which would delete the namespace as well as everything within it,
+			// so we want to avoid using `kubectl replace` in that case as well.
+			if kube.IsCRD(t.targetObj) || t.targetObj.GetKind() == kubeutil.NamespaceKind {
+				update := t.targetObj.DeepCopy()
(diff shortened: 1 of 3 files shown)


pkg/sync/sync_context.go
```diff
@@ -903,57 +903,44 @@ func (sc *syncContext) ensureCRDReady(name string) error {
...
-		// The same thing applies for namespaces, which would delete the namespace as well as everything within it,
-		// so we want to avoid using `kubectl replace` in that case as well.
-		if kube.IsCRD(t.targetObj) || t.targetObj.GetKind() == kubeutil.NamespaceKind {
-			update := t.targetObj.DeepCopy()
-			update.SetResourceVersion(t.liveObj.GetResourceVersion())
-			_, err = sc.resourceOps.UpdateResource(context.TODO(), update, dryRunStrategy)
-			if err != nil {
-				return fmt.Sprintf("error when updating: %v", err.Error()), err
+	// if it is a dry run, disable server side apply, as the goal is to validate only the
+	// yaml correctness of the rendered manifests.
```
leoluz, line 919: Please add in the comment that running dry-run in server mode breaks the auto create namespace feature and link the issue: https://github.com/argoproj/argo-cd/issues/13874

This way future contributors will understand why we are forcing client-side apply for the dry-run phase.

pkg/utils/kube/resource_ops.go
```diff
@@ -229,7 +229,11 @@ func (k *kubectlResourceOperations) ApplyResource(ctx context.Context, obj *unst
 	span.SetBaggageItem("kind", obj.GetKind())
 	span.SetBaggageItem("name", obj.GetName())
 	defer span.Finish()
-	k.log.Info(fmt.Sprintf("Applying resource %s/%s in cluster: %s, namespace: %s", obj.GetKind(), obj.GetName(), k.config.Host, obj.GetNamespace()))
+	k.log.WithValues(
+		"dry-run", dryRunStrategy,
```
leoluz, line 233: Please implement the String function in the DryrunStrategy type so we can have a more meaningful logs.

leoluz (changes requested):
Added some minor comments.
Otherwise LGTM.
Tks a lot for working on this!

leoluz:
@anandf Please fix the DCO check

pkg/sync/sync_context.go
anandf: Updated the comments in both places, with a reference to the actual bug.

pkg/utils/kube/resource_ops.go
anandf: DryRunStrategy  is from kubectl library `k8s.io/kubectl/pkg/cmd/util` . So I get an error when adding String() fn
cannot define new methods on non-local type "k8s.io/kubectl/pkg/cmd/util".DryRunStrategy

Created an inline array lookup as below which is concise.
```
[...]string{"none", "client", "server"}[dryRunStrategy]
```
the messages now look like below for dry runs
```
time="2024-01-19T18:38:21Z" level=info msg="Applying resource Namespace/guestbook in cluster: [https://10.96.0.1:443](https://10.96.0.1/), namespace: " dry-run=client manager=argocd-controller serverSideApply=false serverSideDiff=false
```
and for the actual syncs, it looks like below:
```
time="2024-01-19T18:38:23Z" level=info msg="Applying resource Deployment/kustomize-guestbook-ui in cluster: [https://10.96.0.1:443](https://10.96.0.1/), namespace: guestbook" dry-run=none manager=argocd-controller serverSideApply=true serverSideDiff=false
```

leoluz (approved):
LGTM

PR #569, opened 2024-04-11
fix: temporary files written to '/dev/shm' not cleaned up (#568)

### LINKED ISSUE #568: In 'ManageResources' and 'runResourceCommand', temporary files written to '/dev/shm' may not be cleaned up, in rare cases
We've had user reports of `/dev/shm` running out of space due to temporary files left hanging around within the application-container of Argo CD. Within Argo CD's application-controller, the references to `/dev/shm` are limited to gitops-engine's `TempDir` in `pkg/utils/io`.

I've thus been auditing gitops-engine's use of `/dev/shm`, and discovered there are a couple (rare) cases where the temporary files that are created in `/dev/shm` would not be cleaned up.

https://github.com/argoproj/gitops-engine/pull/569 has been raised to fix the issue.

jgwest:
Fixes #568 

See [parent issue](https://github.com/argoproj/gitops-engine/issues/568) for details of the problem

This PR:
- `ctl.go`: Adds a call to `DeleteFile` to handle the (rare) case where `WriteKubeConfig` fails.
- `resource_ops.go`: Moves `defer io.DeleteFile(...)` call to handle the (rare) case where Write or Close fail.

Admittedly, these are likely to be rare: it's unlikely that writing to a file would fail, when that file was already successfully created earlier in the function.

### DIFF
--- a/pkg/utils/io/io.go
+++ b/pkg/utils/io/io.go
@@ -5,6 +5,7 @@ import (
 )
 
 var (
+	// TempDir is set to '/dev/shm' if exists, otherwise is "", which defaults to os.TempDir() when passed to os.CreateTemp()
 	TempDir string
 )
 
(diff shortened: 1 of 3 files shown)


jannfis (approved):
LGTM

PR #574, opened 2024-05-09
fix: handle nil ParseableType from GVKParser

leoluz:
fix https://github.com/argoproj/argo-cd/issues/18020

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -209,7 +209,7 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
 	gvk := predictedLive.GetObjectKind().GroupVersionKind()
 	pt := gvkParser.Type(gvk)
 	if pt == nil {
-		return nil, fmt.Errorf("unable to resolve parseableType for gvk: %s", gvk)
+		return nil, fmt.Errorf("unable to resolve parseableType for GroupVersionKind: %s", gvk)
 	}
 
 	typedPredictedLive, err := pt.FromUnstructured(predictedLive.Object)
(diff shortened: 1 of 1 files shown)


pkg/diff/diff.go
```diff
@@ -208,6 +208,10 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
 	}
 	gvk := predictedLive.GetObjectKind().GroupVersionKind()
 	pt := gvkParser.Type(gvk)
+	if pt == nil {
+		return nil, fmt.Errorf("unable to resolve parseableType for gvk: %s", gvk)
```
todaywasawesome, line 212:
```suggestion
		return nil, fmt.Errorf("unable to resolve parseableType for gvk (Group Version Kind): %s", gvk)
```
I think we can make this a bit more clear.

pkg/diff/diff.go
```diff
@@ -316,6 +320,9 @@ func structuredMergeDiff(p *SMDParams) (*DiffResult, error) {
 
 	gvk := p.config.GetObjectKind().GroupVersionKind()
 	pt := gescheme.ResolveParseableType(gvk, p.gvkParser)
+	if pt == nil {
+		return nil, fmt.Errorf("unable to resolve parseableType for gvk: %s", gvk)
```
todaywasawesome, line 324:
```suggestion
		return nil, fmt.Errorf("unable to resolve parseableType for gvk (Group Version Kind): %s", gvk)
```
Same

todaywasawesome (changes requested):
If you're cool with it, let's update the text so it's a bit more clear for users.

crenshaw-dev (approved):

pkg/diff/diff.go
leoluz: @todaywasawesome This is an internal error message but I agree with making it a bit more clear. I added a small modification. 
PTAL

pkg/diff/diff.go
leoluz: @todaywasawesome This is an internal error message but I agree with making it a bit more clear. I added a small modification. 
PTAL

todaywasawesome (approved):

PR #577, opened 2024-05-14
feat: Add ability to hide certain annotations on secret resources

svghadi:
Related to https://github.com/argoproj/argo-cd/issues/15693. 

This PR implements core logic from [argoproj/argo-cd#hide-annotations.md](https://github.com/argoproj/argo-cd/blob/master/docs/proposals/feature-bounties/hide-annotations.md) proposal to hide annotations on secret resources. 

This change will be integrated with Argo CD via https://github.com/argoproj/argo-cd/pull/18216. 

_Integration results:_

https://github.com/user-attachments/assets/f47d7b18-cfbc-4fef-8a63-cd7e670abcc7

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -1045,8 +1045,12 @@ func HideSecretData(target *unstructured.Unstructured, live *unstructured.Unstru
 		valToReplacement := make(map[string]string)
 		for _, obj := range []*unstructured.Unstructured{target, live, orig} {
 			var annots map[string]interface{}
+			var err error
 			if obj != nil {
-				annots, _, _ = unstructured.NestedMap(obj.Object, "metadata", "annotations")
+				annots, _, err = unstructured.NestedMap(obj.Object, "metadata", "annotations")
+				if err != nil {
+					return nil, nil, fmt.Errorf("unstructured.NestedMap error: %s", err)
+				}
 			}
 			if annots == nil {
 				annots = make(map[string]interface{})
@@ -1063,7 +1067,10 @@ func HideSecretData(target *unstructured.Unstructured, live *unstructured.Unstru
 				valToReplacement[strVal] = replacement
 			}
 			annots[k] = replacement
-			unstructured.SetNestedMap(obj.Object, annots, "metadata", "annotations")
+			err = unstructured.SetNestedMap(obj.Object, annots, "metadata", "annotations")
+			if err != nil {
+				return nil, nil, fmt.Errorf("unstructured.SetNestedField error: %s", err)
+			}
 		}
 	}
 
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -972,7 +972,7 @@ func CreateTwoWayMergePatch(orig, new, dataStruct interface{}) ([]byte, bool, er
 // HideSecretData replaces secret data & optional annotations values in specified target, live secrets and in last applied configuration of live secret with stars. Also preserves differences between
 // target, live and last applied config values. E.g. if all three are equal the values would be replaced with same number of stars. If all the are different then number of stars
 // in replacement should be different.
-func HideSecretData(target *unstructured.Unstructured, live *unstructured.Unstructured, hideAnnots ...string) (*unstructured.Unstructured, *unstructured.Unstructured, error) {
+func HideSecretData(target *unstructured.Unstructured, live *unstructured.Unstructured, hideAnnots map[string]bool) (*unstructured.Unstructured, *unstructured.Unstructured, error) {
 	var orig *unstructured.Unstructured
 	if live != nil {
 		orig, _ = GetLastAppliedConfigAnnotation(live)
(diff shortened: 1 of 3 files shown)


pasha-codefresh:
@CodiumAI-Agent /review

pkg/diff/diff.go
```diff
@@ -1060,6 +1061,43 @@ func HideSecretData(target *unstructured.Unstructured, live *unstructured.Unstru
 			}
 		}
 	}
+
+	// hide annotations
+	for _, k := range hideAnnots {
```
pasha-codefresh, line 1066: can we move this part to generic function that will hide some values, only differences between your change and the secrets hiding that already exists. 

Only difference is how you fetch "data". I think it can be generic function that accept another function that return "data"

pasha-codefresh (changes requested):

pkg/diff/diff.go
svghadi: Incorporated your suggestion. Moved hide logic to a generic function.

pasha-codefresh: @svghadi will review it tomorrow

pasha-codefresh:
@svghadi sorry, had urgent work to do, i will review it tomorrow

pkg/diff/diff.go
```diff
@@ -969,10 +969,10 @@ func CreateTwoWayMergePatch(orig, new, dataStruct interface{}) ([]byte, bool, er
 	return patch, string(patch) != "{}", nil
 }
 
-// HideSecretData replaces secret data values in specified target, live secrets and in last applied configuration of live secret with stars. Also preserves differences between
+// HideSecretData replaces secret data & optional annotations values in specified target, live secrets and in last applied configuration of live secret with stars. Also preserves differences between
 // target, live and last applied config values. E.g. if all three are equal the values would be replaced with same number of stars. If all the are different then number of stars
 // in replacement should be different.
-func HideSecretData(target *unstructured.Unstructured, live *unstructured.Unstructured) (*unstructured.Unstructured, *unstructured.Unstructured, error) {
+func HideSecretData(target *unstructured.Unstructured, live *unstructured.Unstructured, hideAnnots map[string]bool) (*unstructured.Unstructured, *unstructured.Unstructured, error) {
```
jsoref, line 975: Personally, I don't see any reason to abbreviate `annotations` as `annots`.

(As someone who works on a spell-checker, I get rather annoyed by custom abbreviations -- they make things much harder to deal with.)

pkg/diff/diff.go
```diff
@@ -969,10 +969,10 @@ func CreateTwoWayMergePatch(orig, new, dataStruct interface{}) ([]byte, bool, er
 	return patch, string(patch) != "{}", nil
 }
 
-// HideSecretData replaces secret data values in specified target, live secrets and in last applied configuration of live secret with stars. Also preserves differences between
+// HideSecretData replaces secret data & optional annotations values in specified target, live secrets and in last applied configuration of live secret with stars. Also preserves differences between
 // target, live and last applied config values. E.g. if all three are equal the values would be replaced with same number of stars. If all the are different then number of stars
```
jsoref, line 973: This appears to be a lie.

The code uses `+` (plus) characters not `*` (star) characters.

It's a preexisting condition, but it'd be nice if it were fixed (in a distinct commit).

pkg/diff/diff.go
```diff
@@ -969,10 +969,10 @@ func CreateTwoWayMergePatch(orig, new, dataStruct interface{}) ([]byte, bool, er
 	return patch, string(patch) != "{}", nil
 }
 
-// HideSecretData replaces secret data values in specified target, live secrets and in last applied configuration of live secret with stars. Also preserves differences between
+// HideSecretData replaces secret data & optional annotations values in specified target, live secrets and in last applied configuration of live secret with stars. Also preserves differences between
```
jsoref, line 972: I kinda object to `Hide` as a verb, this appears to be closer to `Mask` than `Hide`.

That appears to be a preexisting condition, but if I inherited this code, I'd want to change it.

pkg/diff/diff.go
svghadi: Ack. Updated it to `hideAnnotations`.

pkg/diff/diff.go
svghadi: Updated the comment to reflect this.

pkg/diff/diff_test.go
```diff
@@ -1006,13 +1010,141 @@ func TestHideSecretDataSameKeysSameValues(t *testing.T) {
...
+			name:           "no hidden annotations",
+			hideAnnots:     nil,
+			annots:         map[string]interface{}{"token/value": "secret", "key": "secret-key", "app": "test"},
+			expectedAnnots: map[string]interface{}{"token/value": "secret", "key": "secret-key", "app": "test"},
+		},
+		{
+			name:           "hide annotations",
+			hideAnnots:     map[string]bool{"token/value": true, "key": true},
+			annots:         map[string]interface{}{"token/value": "secret", "key": "secret-key", "app": "test"},
+			expectedAnnots: map[string]interface{}{"token/value": replacement1, "key": replacement1, "app": "test"},
```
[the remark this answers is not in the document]
svghadi: AFAIU, the number of `+` symbols changes only when the value for a particular key differs between the live and target objects. There is no connection between the values of different keys.

pkg/diff/diff.go
```diff
@@ -995,6 +995,36 @@ func HideSecretData(target *unstructured.Unstructured, live *unstructured.Unstru
...
+		if err != nil {
+			return nil, nil, fmt.Errorf("error marshaling json: %s", err)
+		}
+		annotations[corev1.LastAppliedConfigAnnotation] = string(lastAppliedData)
+		live.SetAnnotations(annotations)
+	}
+	return target, live, nil
+}
+
+func hide(target, live, orig *unstructured.Unstructured, keys map[string]bool, fields ...string) (*unstructured.Unstructured, *unstructured.Unstructured, *unstructured.Unstructured, error) {
```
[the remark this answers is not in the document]
svghadi: Thanks @agaudreault for pointing out a edge case for `lastAppliedAnnotation`.  I have handled it and also incorporated your suggestion of renaming `orig` to `liveLastAppliedAnnotation`.

pkg/diff/diff.go
```diff
@@ -995,25 +996,58 @@ func HideSecretData(target *unstructured.Unstructured, live *unstructured.Unstru
...
 			if obj != nil {
 				// handles an edge case when secret data has nil value
 				// https://github.com/argoproj/argo-cd/issues/5584
-				dataValue, ok := obj.Object["data"]
-				if ok {
-					if dataValue == nil {
-						continue
-					}
+				dataValue, ok, _ := unstructured.NestedFieldCopy(obj.Object, fields...)
+				if !ok || dataValue == nil {
```
pasha-codefresh, line 1044: I am not sure that it was not bug before, but you changed logic a bit. Now if ok will be false, it will execute continue, before it just skipped this condition.

pasha-codefresh (changes requested):

pkg/diff/diff.go
```diff
@@ -995,25 +996,58 @@ func HideSecretData(target *unstructured.Unstructured, live *unstructured.Unstru
...
+
+	// hide last-applied-config annotation
+	if live != nil && liveLastAppliedAnnotation != nil {
+		annotations := live.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		// special case: hide "kubectl.kubernetes.io/last-applied-configuration" annotation
+		if _, ok := hideAnnotations[corev1.LastAppliedConfigAnnotation]; ok {
+			annotations[corev1.LastAppliedConfigAnnotation] = replacement
```
pasha-codefresh, line 1020: why we hide whole annotation if we already hide secrets during previous hide() calls ?

pasha-codefresh (changes requested):

pkg/diff/diff.go
svghadi: A user might be interested in hiding the whole `last-applied-config` annotation by setting the key in configmap. Without this special case logic, `last-applied-config` annotation will never be hidden if user wants to hide it completely.

pasha-codefresh: understand, thank you

pkg/diff/diff.go
svghadi: Yes. I think, I made a mistake while trying to simplify the nested if. I have reverted it to the old conditions. Thanks for catching it.

agaudreault (approved):
LGTM

pasha-codefresh:
Taking one more day for manual testing, but overall LGTM to me

pasha-codefresh (approved):
LGTM, thank you a lot

PR #588, opened 2024-06-27
chore: update static scheme

crenshaw-dev:
Should fix the following:

https://github.com/argoproj/argo-cd/issues/13585
https://github.com/argoproj/argo-cd/issues/15176
https://github.com/argoproj/argo-cd/issues/16640
https://github.com/argoproj/argo-cd/issues/17358
https://github.com/argoproj/argo-cd/issues/17548
https://github.com/argoproj/argo-cd/issues/17811
https://github.com/argoproj/argo-cd/issues/18778

I've tested most of them manually. I don't think unit tests would be valuable. We can test old fields but not new ones.


leoluz (approved):
LGTM
Tks!

PR #587, opened 2024-06-25
fix: deduplicate OpenAPI definitions for GVKParser

crenshaw-dev:
Argo CD PR: https://github.com/argoproj/argo-cd/pull/18840

PR #592, opened 2024-07-01
fix: deduplicate OpenAPI definitions for GVKParser (#587) (#589) (2.10)

crenshaw-dev:
* fix: deduplicate OpenAPI definitions for GVKParser

* do the thing that was the whole point

* more logs

* don't uniquify models

* schema for both

* more logs

* fix logic

* better tainted gvk handling, better docs, update mocks

* add a test

* improvements from comments

---------


leoluz (approved):
LGTM

PR #601, opened 2024-07-04
feat: More optimal IterateHierarchyV2 and iterateChildrenV2 [#600]

### LINKED ISSUE #600: Implement more optimal versions of `IterateHierarchy` and `iterateChildren` based on nodes graph
These should have a linear time complexity instead of up to quadratic. See https://github.com/argoproj/argo-cd/issues/18929 for more details.

andrii-korotkov-verkada:
Closes #600

The existing (effectively v1) implementations are suboptimal since they don't construct a graph before the iteration. They search for children by looking at all namespace resources and checking `isParentOf`, which can give `O(tree_size * namespace_resources_count)` time complexity. The v2 algorithms construct the graph and have `O(namespace_resources_count)` time complexity. See more details in the linked issues.


andrii-korotkov-verkada:
Testing with ArgoCD https://github.com/argoproj/argo-cd/pull/18972

andrii-korotkov-verkada:
Looks really good on live cluster. ~300ms instead of almost ~4m for the same application!
Screenshot 2024-07-07 at 4 29 56 PM

andrii-korotkov-verkada:
Here are some perf views of the system collected following https://github.com/argoproj/argo-cd/issues/13534#issuecomment-1562184630.

The build is from master on 2024/07/07 including https://github.com/argoproj/argo-cd/pull/18972, https://github.com/argoproj/argo-cd/pull/18694, https://github.com/argoproj/gitops-engine/pull/601, https://github.com/argoproj/gitops-engine/pull/603.

Screenshot 2024-07-09 at 3 47 18 PM
Screenshot 2024-07-09 at 3 48 02 PM
Screenshot 2024-07-09 at 3 48 39 PM

pkg/cache/cluster.go
```diff
@@ -1000,6 +1003,94 @@ func (c *clusterCache) IterateHierarchy(key kube.ResourceKey, action func(resour
...
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	keysPerNamespace := make(map[string][]kube.ResourceKey)
+	for _, key := range keys {
+		keysPerNamespace[key.Namespace] = append(keysPerNamespace[key.Namespace], key)
+	}
+	for namespace, namespaceKeys := range keysPerNamespace {
+		nsNodes := c.nsIndex[namespace]
+		// Prepare to construct a graph
+		nodesByUID := make(map[types.UID][]*Resource)
```
crenshaw-dev, line 1017: Just trying to understand this data structure... how is it possible for multiple nodes to have the same UID?

crenshaw-dev: Ah okay, I'm seeing this from iterate V1:

```
				// The object might have multiple children with the same UID (e.g. replicaset from apps and extensions group). It is ok to pick any object but we need to make sure
				// we pick the same child after every refresh.
```

andrii-korotkov-verkada: Yes, I just tried to preserve quirks of existing logic.

pkg/cache/cluster.go
```diff
@@ -1004,6 +1007,94 @@ func (c *clusterCache) IterateHierarchy(key kube.ResourceKey, action func(resour
 	}
 }
 
+// IterateHierarchy iterates resource tree starting from the specified top level resources and executes callback for each resource in the tree
+func (c *clusterCache) IterateHierarchyV2(keys []kube.ResourceKey, action func(resource *Resource, namespaceResources map[kube.ResourceKey]*Resource) bool) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	keysPerNamespace := make(map[string][]kube.ResourceKey)
+	for _, key := range keys {
+		keysPerNamespace[key.Namespace] = append(keysPerNamespace[key.Namespace], key)
```
crenshaw-dev, line 1016: The existing `IterateHierarchy` method short-circuits the whole function if the key doesn't exist in `c.resources`. We should keep that optimization here.

andrii-korotkov-verkada: I have a similar block below
```
	for _, key := range namespaceKeys {
			res, ok := c.resources[key]
			if !ok {
				continue
			}
```
But yeah, I guess I can skip it here.

andrii-korotkov-verkada: Updated the PR, hopefully addressing this.

crenshaw-dev: Yep, that check was after building the graph, I think we should do the check before.

crenshaw-dev: Update looks good to me!

pkg/cache/cluster.go
```diff
@@ -1004,6 +1007,98 @@ func (c *clusterCache) IterateHierarchy(key kube.ResourceKey, action func(resour
...
+		}
+		visited := make(map[kube.ResourceKey]int)
+		for _, key := range namespaceKeys {
+			visited[key] = 0
+		}
+		for _, key := range namespaceKeys {
+			res, ok := c.resources[key]
+			if !ok {
+				continue
+			}
```
crenshaw-dev, line 1070: This check is redundant now, right?

andrii-korotkov-verkada: Yeah, updated the PR to remove it.

crenshaw-dev:
Could you take a look at this PR? https://github.com/andrii-korotkov-verkada/gitops-engine/pull/1

I'll rebase, the force-push messed up the diff.

crenshaw-dev:
Pushed.

andrii-korotkov-verkada:
@crenshaw-dev, looks good! Thanks for improving the code. I've merged the changes.

crenshaw-dev:
@andrii-korotkov-verkada thanks! I may have a few more as I continue digging through the code. Bear with me. I plan to stick with it this week until we get it merged, as long as you have time to keep working on it!

crenshaw-dev:
More fun: https://github.com/andrii-korotkov-verkada/gitops-engine/pull/2

andrii-korotkov-verkada:
Sounds good! Thanks for the help. I'd be pretty active on this.

pkg/cache/cluster.go
```diff
@@ -1004,6 +1007,109 @@ func (c *clusterCache) IterateHierarchy(key kube.ResourceKey, action func(resour
...
+type graphKey struct {
+	kind       string
+	apiVersion string
+	name       string
+}
+
+func buildGraph(nsNodes map[kube.ResourceKey]*Resource) (map[kube.ResourceKey][]kube.ResourceKey, map[kube.ResourceKey]map[types.UID]*Resource) {
+	// Prepare to construct a graph
+	nodesByUID := make(map[types.UID][]*Resource, len(nsNodes))
+	nodeByGraphKey := make(map[graphKey]*Resource, len(nsNodes))
```
crenshaw-dev, line 1061: Do we really need `nodeByGraphKey`, or could we just use `nsNodes`, since all the resources should be in the same namespace?

crenshaw-dev: Using nsNodes passes gitops-engine unit tests, but I'm skeptical. I can put up a PoC tomorrow to analyze.

andrii-korotkov-verkada: `nodeByGraphKey` is for efficient node lookup during uid backfill. I don't know if we avoid it.

andrii-korotkov-verkada: It's slightly different comparing to the kube resource key, e.g. uses api version instead of group.

andrii-korotkov-verkada: I don't know the original reasoning for this distinction, but left it for backwards compatibility.

crenshaw-dev:
Probably the last graph building optimization: https://github.com/andrii-korotkov-verkada/gitops-engine/pull/3

crenshaw-dev:
@andrii-korotkov-verkada up for considering this one? https://github.com/andrii-korotkov-verkada/gitops-engine/pull/4

I think we do risk missing parent relationships due to Resources lacking the correct namespace field. But so far unit tests, Argo CD e2e tests, and running in an Intuit system looks okay.

I think it would be worth the risk for cutting the memory use in half and saving ~30% execution time.

andrii-korotkov-verkada:
In general it'd make sense to me since all other places use group. I'll merge it.

crenshaw-dev:
Test results from Intuit:

Before, IterateHierarchy was taking about 10% of the application controller's time while refreshing apps. Now it takes around 1%. So roughly 6s before, now less than 1s out of a 60s profile.

Heap use has gone up by about 2x, but it wasn't memory-heavy before, and it hasn't really increased in a problematic way.

Steady-state CPU and memory is basically the same as before.

I used log metrics to measure 95th percentile reconciliation, getting the resource tree, and setting managed resources.

Reconciliation time is about the same, maaaaybe 25% faster. Getting the resource tree is about the same amount of time. Setting the managed resources now takes ~one fifth the time it did before, down to 200ms from 1000ms.

I think the vast majority of that improvement doesn't actually come from the graph pre-build optimization, but instead comes from avoiding iterating over all the managed resources (an optimization which doesn't actually depend on pre-building the graph).

In summary: my test showed no performance regressions, no functionality regressions, and maybe a slight performance improvement related to the new algorithm. I suspect the relatively small improvement is because we have relatively few resources per-namespace. Bigger namespaces will have a bigger CPU win and (probably) a higher memory cost.

crenshaw-dev (approved):

andrii-korotkov-verkada:
Thanks for all the testing and support! Yeah, I guess we just have quite large default namespace (which isn't best practice, but here we are), hence I saw larger wins. Longest running processor in the largest cluster went down from 30-60min to 1-2min.

PR #605, opened 2024-07-11
fix: swallowed error in startMissingWatches

agaudreault:
If loadInitialState returns an error, it is possible to still call watchEvents and loadInitialState again

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -492,21 +492,24 @@ func (c *clusterCache) startMissingWatches() error {
 
 			err := c.processApi(client, api, func(resClient dynamic.ResourceInterface, ns string) error {
 				resourceVersion, err := c.loadInitialState(ctx, api, resClient, ns, false) // don't lock here, we are already in a lock before startMissingWatches is called inside watchEvents
-				if err != nil && c.isRestrictedResource(err) {
-					keep := false
-					if c.respectRBAC == RespectRbacStrict {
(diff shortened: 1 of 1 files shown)


pkg/cache/cluster.go
```diff
@@ -492,21 +492,24 @@ func (c *clusterCache) startMissingWatches() error {
...
 						}
-						keep = k
-					}
-					// if we are not allowed to list the resource, remove it from the watch list
-					if !keep {
-						delete(c.apisMeta, api.GroupKind)
-						delete(namespacedResources, api.GroupKind)
-						return nil
 					}
+					return fmt.Errorf("failed to start watch %s: %w", api.GroupKind.String(), err)
```
crenshaw-dev, line 512: Shouldn't this be wrapped in an `else` to avoid throwing an error when the resource is just restricted?

agaudreault: I dont think it makes sense to call `c.watchEvents` if resourceVersion is nil and there is an error. 

If keep is true, that means the resource is restricted `(k8sErrors.IsForbidden(err) || k8sErrors.IsUnauthorized(err))`, but the controller has permissions... it sounds like it should never happens.

And there is another codepath here that has a a different logic, so I think both should be the same. @gdsoumya can you give more insights on what should be happening.
https://github.com/argoproj/gitops-engine/blob/b04ee0458ad719cf806bc2c5d40420364b99b552/pkg/cache/cluster.go#L870-L871

PR #612, opened 2024-07-16
feat: add ExternalSecret to sync wave

ckav370:
I want to be able to use a `preSync` hook to run some database migrations. This has been working fine, until such a time as updating the container secrets which are using the External Secret Operator and CRD to create a k8s secret which the job created by the `preSync` hook creates mounts. 

As per this Github issue [#9891](https://github.com/argoproj/argo-cd/issues/9891), the recommended approach is to use Argo sync waves. However `ExternalSecret` is not included in one of the predefined `kinds` and therefore cannot be part of a different "wave" meaning that any secrets defined by `ExternalSecrets` cannot be synced before the `preSync` hook as part of the pre Sync wave.

This would allow the `ExternalSecret` to have an annotation such as the following
```
argocd.argoproj.io/sync-wave: "-2"
```

And a `preSync` hook to have a greater weighted sync wave such as 
```
argocd.argoproj.io/sync-wave: "-1"

```
And all other resources which are part of the deployment would default to weight of `0` without the need for an annotation

### DIFF
--- a/pkg/sync/sync_tasks.go
+++ b/pkg/sync/sync_tasks.go
@@ -60,6 +60,7 @@ func init() {
 		"IngressClass",
 		"Ingress",
 		"APIService",
+		"ExternalSecrets",
 	}
 	for i, kind := range kinds {
 		// make sure none of the above entries are zero, we need that for custom resources
--- a/pkg/sync/sync_tasks_test.go
+++ b/pkg/sync/sync_tasks_test.go
@@ -13,8 +13,8 @@ import (
 )
 
 func Test_syncTasks_kindOrder(t *testing.T) {
-	assert.Equal(t, -35, kindOrder["Namespace"])
-	assert.Equal(t, -1, kindOrder["APIService"])
+	assert.Equal(t, -36, kindOrder["Namespace"])
+	assert.Equal(t, -2, kindOrder["APIService"])
 	assert.Equal(t, 0, kindOrder["MyCRD"])
 }
 


schlags:
Isn't this already possible?

```yaml
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  annotations:
    argocd.argoproj.io/hook: PreSync
    argocd.argoproj.io/sync-wave: "-1000"
```

We already use ExternalSecrets in this fashion. One of our PreSync jobs requires a secret present and this was our strategy to prevent the catch 22 of initial sync not having that secret defined yet.

kostis-codefresh:
Hello

First of all, looking at the commit on its own I have two comments

1) The list you modify contains a list of built-in Kubernetes resources. What you are adding is not a built-in resource. Even if that would fix your use case, the next person would then add their own custom CRD and so on. This is not sustainable in the long run (for everybody to add their own resource there)
2) As the comment says this list is designed on purpose to mimic the [default ordering that Helm follows](https://github.com/helm/helm/blob/0361dc85689e3a6d802c444e2540c92cb5842bc9/pkg/releaseutil/kind_sorter.go#L31). By editing this list, you break this convention. 

However just to take a step backwards. Can you explain what exactly is your use case and why it is not covered by sync waves? And I mean JUST sync waves and not phases/hooks.

> However ExternalSecret is not included in one of the predefined kinds and therefore cannot be part of a different "wave" meaning that any secrets defined by ExternalSecrets cannot be synced before the preSync hook as part of the pre Sync wave.

I don't understand this sentence. Could you please attach a minimal example of two resources with just sync waves and explain what Argo CD does versus what you expected to happen?

alexrecuenco:
I think they are trying to achieve this by forcing the ordering as a default. 

> https://github.com/argoproj/argo-cd/issues/9891#issuecomment-2090435031

It would be great to be able to set some default ordering sometimes through a configmap that apply globally to solve the use case here. 

i.e., "In my cluster, every ExternalSecret has to be synced in sync wave X before continuing"

piotrplenik:
I would treat this issue more as a bug than a feature: if we use `ExternalSecret`, there is no way to keep proper order and prevent deletion.

The second thing is that more resources should be added:
 - `SecretStore`
 - `ClusterSecretStore`
 - `ClusterExternalSecret`
 - `PushSecret`

ref https://external-secrets.io/latest/api/externalsecret/

pkg/sync/sync_tasks.go
```diff
@@ -60,6 +60,7 @@ func init() {
 		"IngressClass",
 		"Ingress",
 		"APIService",
+		"ExternalSecrets",
```
piotrplenik, line 63:
```suggestion
		"SecretStore",
		"ClusterSecretStore",
		"ClusterExternalSecret",
		"PushSecret",
		"ExternalSecrets",
```

PR #613, opened 2024-07-17
Enable consumers of gitops-engine API to specify an alternative temp directory for generated manifests, as an alternative '/dev/shm'

jgwest:
See parent issue for details: https://github.com/argoproj/gitops-engine/issues/614

This PR:
- When calling gitops-engine, you will now need to explicitly define the temporary path to write kubectl manifests to:
	- Add new parameter `tmpManifestPath` to public APIs: 
		- `NewEngine` in `pkg/engine/engine.go`
		- `NewClusterCache` in `pkg/cache/cluster.go
	- In both cases, when calling these functions, you now must specify where to store temporary manifest files that are generated
	- If you wish to preserve the existing behaviour, just use the value from calling `io.TempPathUseDevShmIfAvailable()` (see examples in this PR)
- Updates existing gitops-engine code to use new API
- Preserves the existing behaviour of the code: code and tests that were previously using `/dev/shm` will continue to do so
- Updates `pkg/utils/io/io.go` API to make it more explicit about behaviour

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/klog/v2/textlogger"
 	"k8s.io/kubectl/pkg/util/openapi"
 
+	"github.com/argoproj/gitops-engine/pkg/utils/io"
 	"github.com/argoproj/gitops-engine/pkg/utils/kube"
 	"github.com/argoproj/gitops-engine/pkg/utils/tracing"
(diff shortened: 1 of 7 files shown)


ppapapetrou76:
The gitops-engine repository is migrating to https://github.com/argoproj/argo-cd. We are closing all draft PRs.
Please feel free to submit it again once the migration is over.

PR #625, opened 2024-09-10
chore: speed up resolveResourceReferences

crenshaw-dev:
Sitting at the airport with nothing better to do than micro-optimize stuff.

```
before: Benchmark_resolveResourceReferences-16           4676506               252.4 ns/op            82 B/op          3 allocs/op
after: Benchmark_resolveResourceReferences-16            8308146               143.6 ns/op            82 B/op          3 allocs/op
```

I didn't commit the benchmark code, because it's trivial and shouldn't be needed in the future. Here it is for inspection:

```go
func Benchmark_resolveResourceReferences(b *testing.B) {
	manifest := []byte(`
apiVersion: extensions/v1beta2
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    foo: bar
  uid: '123'
  ownerReferences:
  - apiVersion: apps/v1
    kind: StatefulSet
    name: sw-broker
    uid: '456'
`)
	obj := unstructured.Unstructured{}
	err := yaml.Unmarshal(manifest, &obj.Object)
	require.NoError(b, err)

	c := clusterCache{}

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		c.resolveResourceReferences(&obj)
	}
}
```

### DIFF
--- a/pkg/cache/references.go
+++ b/pkg/cache/references.go
@@ -3,11 +3,11 @@ package cache
 import (
 	"encoding/json"
 	"fmt"
-        "regexp"
 	v1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
+	"regexp"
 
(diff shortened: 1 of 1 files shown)


agaudreault (approved):

crenshaw-dev:
Argo CD tests passed, merging. Thanks, Alex!

PR #628, opened 2024-09-16
chore: avoid unnecessary deep copy

crenshaw-dev:
I'm not convinced we need to deep copy these objects at all. But to be cautious, I think we can save some resources by only deep copying if we need to mutate the objects.

Here's the benchmark result:

```
before: Benchmark_threeWayMergePatch-16            23341             52004 ns/op           39070 B/op        810 allocs/op
after:  Benchmark_threeWayMergePatch-16            24838             48706 ns/op           37709 B/op        800 allocs/op
```

We save a little memory and CPU.

Here's the benchmark code:

```go
func Benchmark_threeWayMergePatch(b *testing.B) {
	orig := []byte(`
apiVersion: v1
kind: Service
metadata:
  name: my-service
  namespace: default
`)
	config := []byte(`
apiVersion: v1
kind: Service
metadata:
  name: my-service
  namespace: default
`)
	live := []byte(`
apiVersion: v1
kind: Service
metadata:
  name: my-service
  namespace: default
`)

	origUnstructured := unstructured.Unstructured{}
	err := yaml.Unmarshal(orig, &origUnstructured)
	require.NoError(b, err)
	configUnstructured := unstructured.Unstructured{}
	err = yaml.Unmarshal(config, &configUnstructured)
	require.NoError(b, err)
	liveUnstructured := unstructured.Unstructured{}
	err = yaml.Unmarshal(live, &liveUnstructured)
	require.NoError(b, err)

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		_, err := ThreeWayDiff(&origUnstructured, &configUnstructured, &liveUnstructured)
		require.NoError(b, err)
	}
}
```

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -694,21 +694,33 @@ func stripTypeInformation(un *unstructured.Unstructured) *unstructured.Unstructu
 // config or last-applied. This results in a diff which we don't care about. We delete the two so
 // that the diff is more relevant.
 func removeNamespaceAnnotation(orig *unstructured.Unstructured) *unstructured.Unstructured {
-	orig = orig.DeepCopy()
 	if metadataIf, ok := orig.Object["metadata"]; ok {
 		metadata := metadataIf.(map[string]interface{})
-		delete(metadata, "namespace")
+
+		var deleteNamespace, deleteAnnotations bool
+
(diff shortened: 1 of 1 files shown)


crenshaw-dev:
Fixing the test failure makes the change more costly than the original. Leaving it as-is for now.

PR #629, opened 2024-10-04
fix: avoid resources lock contention utilizing channel

mpelekh:
Problem statement is in https://github.com/argoproj/argo-cd/issues/8172#issuecomment-2277585238

The [IterateHierrchyV2](https://github.com/argoproj/gitops-engine/commit/6b2984ebc47085852a7b63a0fd0b73c52e986217) significantly improved performance, getting us ~90% of the way there. **But on huge clusters, we still have significant lock contention.**

The fix in this pull request approaches the problem differently - it avoids lock contention by utilizing a channel to process events from the cluster.

More details are in the comments.

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -200,10 +200,11 @@ func NewClusterCache(config *rest.Config, opts ...UpdateSettingsFunc) *clusterCa
 type clusterCache struct {
 	syncStatus clusterCacheSync
 
-	apisMeta      map[schema.GroupKind]*apiMeta
-	eventMetaCh   chan eventMeta
-	serverVersion string
-	apiResources  []kube.APIResourceInfo
+	apisMeta              map[schema.GroupKind]*apiMeta
+	batchEventsProcessing bool
+	eventMetaCh           chan eventMeta
+	serverVersion         string
+	apiResources          []kube.APIResourceInfo
 	// namespacedResources is a simple map which indicates a groupKind is namespaced
 	namespacedResources map[schema.GroupKind]bool
 
@@ -483,7 +484,9 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
 		opts[i](c)
 	}
 
-	c.invalidateEventMeta()
+	if c.batchEventsProcessing {
+		c.invalidateEventMeta()
+	}
 	c.apisMeta = nil
 	c.namespacedResources = nil
 	c.log.Info("Invalidated cluster")
@@ -867,13 +870,17 @@ func (c *clusterCache) sync() error {
 	for i := range c.apisMeta {
 		c.apisMeta[i].watchCancel()
 	}
-	c.invalidateEventMeta()
+
+	if c.batchEventsProcessing {
+		c.invalidateEventMeta()
+		c.eventMetaCh = make(chan eventMeta)
+	}
+
 	c.apisMeta = make(map[schema.GroupKind]*apiMeta)
 	c.resources = make(map[kube.ResourceKey]*Resource)
 	c.namespacedResources = make(map[schema.GroupKind]bool)
 	config := c.config
 	version, err := c.kubectl.GetServerVersion(config)
-	c.eventMetaCh = make(chan eventMeta)
 
 	if err != nil {
 		return err
@@ -910,7 +917,9 @@ func (c *clusterCache) sync() error {
 		return err
 	}
 
-	go c.processEvents()
+	if c.batchEventsProcessing {
+		go c.processEvents()
+	}
 
 	// Each API is processed in parallel, so we need to take out a lock when we update clusterCache fields.
 	lock := sync.Mutex{}
@@ -1296,7 +1305,13 @@ func (c *clusterCache) recordEvent(event watch.EventType, un *unstructured.Unstr
 		return
 	}
 
-	c.eventMetaCh <- eventMeta{event, un}
+	if c.batchEventsProcessing {
+		c.eventMetaCh <- eventMeta{event, un}
+	} else {
+		c.lock.Lock()
+		defer c.lock.Unlock()
+		c.processEvent(key, eventMeta{event, un})
+	}
 }
 
 func (c *clusterCache) processEvents() {
@@ -1342,19 +1357,23 @@ func (c *clusterCache) processEventsBatch(eventMetas []eventMeta) {
 		}
 	}()
 
-	for _, em := range eventMetas {
-		key := kube.GetResourceKey(em.un)
-		existingNode, exists := c.resources[key]
-		if em.event == watch.Deleted {
-			if exists {
-				c.onNodeRemoved(key)
-			}
-		} else {
-			c.onNodeUpdated(existingNode, c.newResource(em.un))
-		}
+	for _, evMeta := range eventMetas {
+		key := kube.GetResourceKey(evMeta.un)
+		c.processEvent(key, evMeta)
 	}
 
 	log.V(1).Info("Processed events (ms)", "count", len(eventMetas), "duration", time.Since(start).Milliseconds())
+}
+
+func (c *clusterCache) processEvent(key kube.ResourceKey, evMeta eventMeta) {
+	existingNode, exists := c.resources[key]
+	if evMeta.event == watch.Deleted {
+		if exists {
+			c.onNodeRemoved(key)
+		}
+	} else {
+		c.onNodeUpdated(existingNode, c.newResource(evMeta.un))
+	}
 }
 
 func (c *clusterCache) onNodeUpdated(oldRes *Resource, newRes *Resource) {
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -57,6 +57,8 @@ const (
 	// Limit is required to avoid memory spikes during cache initialization.
 	// The default limit of 50 is chosen based on experiments.
 	defaultListSemaphoreWeight = 50
+	// defaultEventProcessingInterval is the default interval for processing events
+	defaultEventProcessingInterval = 1 * time.Second
 )
 
 const (
@@ -73,6 +75,11 @@ type apiMeta struct {
 	// watchCancel stops the watch of all resources for this API. This gets called when the cache is invalidated or when
 	// the watched API ceases to exist (e.g. a CRD gets deleted).
 	watchCancel context.CancelFunc
+}
+
+type eventMeta struct {
+	event watch.EventType
+	un    *unstructured.Unstructured
 }
 
 // ClusterInfo holds cluster cache stats
@@ -95,6 +102,9 @@ type ClusterInfo struct {
 
 // OnEventHandler is a function that handles Kubernetes event
 type OnEventHandler func(event watch.EventType, un *unstructured.Unstructured)
+
+// OnProcessEventsHandler handles process events event
+type OnProcessEventsHandler func(duration time.Duration, processedEventsNumber int)
 
 // OnPopulateResourceInfoHandler returns additional resource metadata that should be stored in cache
 type OnPopulateResourceInfoHandler func(un *unstructured.Unstructured, isRoot bool) (info interface{}, cacheManifest bool)
@@ -137,6 +147,8 @@ type ClusterCache interface {
 	OnResourceUpdated(handler OnResourceUpdatedHandler) Unsubscribe
 	// OnEvent register event handler that is executed every time when new K8S event received
 	OnEvent(handler OnEventHandler) Unsubscribe
+	// OnProcessEventsHandler register event handler that is executed every time when events were processed
+	OnProcessEventsHandler(handler OnProcessEventsHandler) Unsubscribe
 }
 
 type WeightedSemaphore interface {
@@ -153,6 +165,7 @@ func NewClusterCache(config *rest.Config, opts ...UpdateSettingsFunc) *clusterCa
 	cache := &clusterCache{
 		settings:           Settings{ResourceHealthOverride: &noopSettings{}, ResourcesFilter: &noopSettings{}},
 		apisMeta:           make(map[schema.GroupKind]*apiMeta),
+		eventMetaCh:        nil,
 		listPageSize:       defaultListPageSize,
 		listPageBufferSize: defaultListPageBufferSize,
 		listSemaphore:      semaphore.NewWeighted(defaultListSemaphoreWeight),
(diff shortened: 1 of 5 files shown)


mpelekh:
**The issue**

In large clusters where Argo CD monitors numerous resources, the processing of watches becomes significantly slow—in our case (total k8s resources in cluster: ~400k, Pods: ~76k, ReplicaSets: ~52k), taking around 10 minutes. As a result, the Argo CD UI displays outdated information, impacting several features reliant on sync waves, like PruneLast. Eventually, the sheer volume of events from the cluster overwhelmed the system, causing Argo CD to stall completely.

To address this, we disabled the tracking of Pods and ReplicaSets, although this compromises one of the main benefits of the Argo CD UI. We also filtered out irrelevant events and tried to optimize various settings in the application controller. However, vertical scaling of the application controller had no effect, and horizontal scaling is not an option for a single cluster due to sharding limitations.

**Issue causes**

During the issue investigation, it was found that the problem lies in the following:
- resource lock contention
- slow performance of IterateHierarchy

**Patched v2.10.9**

v2.10.9 was patched with the following commits.
- https://github.com/argoproj/gitops-engine/commit/6b2984ebc47085852a7b63a0fd0b73c52e986217
- https://github.com/argoproj/argo-cd/commit/267f243a899483fea0a4e6a613c18f62bd342c7e

<details>
<summary> Though patches significantly improve performance, Argo CD still can not handle the load from large clusters.</summary>


In the screenshot, you can see one of the largest clusters. Here, the patched with the above commits v2.10.9 build is running.
- till 12:50, pods and replica sets are disabled from tracking
- from 12:50 to 13:34, pods and replica sets are enabled to be tracked
- after 13:34, pods and replica sets are disabled from tracking

As can be seen, once pods and rs are enabled to be tracked, the cluster event count falls close to zero, and reconciliation time increases drastically.

Screenshot 2024-08-09 at 20 40 44

Screenshot 2024-08-09 at 20 51 00

Number of pods in cluster: ~76k
Number of rs in cluster: ~52k

A more detailed comparison of different patched versions is added to this comment - https://github.com/argoproj/argo-cd/issues/8172#issuecomment-2278663389

</details>

<details>
<summary>The potential reason is lock contention.</summary>

Here, a few more metrics were added, and it was found that when the number of events is significant, sometimes it takes ~5 minutes to acquire a lock, which leads to a delay in reconciliation.
https://github.com/mpelekh/gitops-engine/commit/560ef00bcce9201083200f906f15bf1716fbfcc0#diff-9c9e197d543705f08c9b1bc2dc404a55506cfc2935a988e6007d248257aadb1aR1372

Screenshot 2024-08-09 at 21 11 33
</details>

The suggested fix https://github.com/argoproj/gitops-engine/issues/602 to optimize the lock usage has not improved the issue in large clusters.


**Avoid resources lock contention utilizing channel**

Since we still have significant lock contention in massive clusters, and the approaches above didn’t resolve the issue, another approach has been considered. It is a part of this PR.

When we must acquire a write lock in each goroutine, we can’t handle more than one event at a time. What if we introduce the channel where all the received events are sent, and one goroutine is responsible for processing events in batch from the channel? In such a way, the locks from each goroutine are moved to the goroutine, which processes events from the channel. This means we would have only one place where the write lock is acquired; in such a way, we would get rid of the lock contention.

<details>
<summary>The fix results</summary>


2a758ca7-9f86-4c98-b0c7-4bf81e6b91e7


As can be seen from metrics, once the fixed version was deployed and Node, ReplicaSets, and Pods were enabled for tracking, the number of cluster events was stable and didn’t go down.

</details>

**Conclusions**
The fix shows significant performance improvements. We left Nodes, ReplicaSets, and Pods enabled on large clusters.
ArgoCD UI is working smoothly.
The original issue has been resolved - users can manage Pods and ReplicaSets on large clusters.

crenshaw-dev:
Your analysis is excruciatingly thorough, I love it! I've posted it to SIG Scalability, and we'll start analyzing ASAP. Please be patient, it'll take us a while to give it a really thorough review.

pkg/cache/cluster.go
```diff
@@ -864,6 +910,8 @@ func (c *clusterCache) sync() error {
 		return err
 	}
 
+	go c.processEvents()
```
crenshaw-dev, line 913: I see this in the `sync` function comment:

```go
// When this function exits, the cluster cache is up to date, and the appropriate resources are being watched for
// changes.
```

If I understand this change correctly (and the associated test changes), by processing these events in a goroutine, we're breaking the guarantee that `sync` will completely update the cluster cache. Is that correct?

crenshaw-dev:
@mpelekh would you be interested in joining a SIG Scalability meeting to talk through the changes?

pkg/cache/cluster.go
mpelekh: These changes do not break the guarantee that `sync` will completely update the cluster cache.
`sync` function populates the cluster cache when it's run.
 https://github.com/argoproj/gitops-engine/blob/a16e663ae6b5a5b0b904175e3672439c54ca7147/pkg/cache/cluster.go#L934

`processEvents` goroutine processes the future events that are received in `watchEvents` goroutine.
https://github.com/argoproj/gitops-engine/blob/a16e663ae6b5a5b0b904175e3672439c54ca7147/pkg/cache/cluster.go#L962

`watchEvents` goroutine watches for the events from k8s resource types. Once an event is received, it's processed.
https://github.com/argoproj/gitops-engine/blob/a16e663ae6b5a5b0b904175e3672439c54ca7147/pkg/cache/cluster.go#L706

The event is sent to the channel that is read in the `processEvents` goroutine, where the processing is done in bulk.
https://github.com/argoproj/gitops-engine/blob/a16e663ae6b5a5b0b904175e3672439c54ca7147/pkg/cache/cluster.go#L1299

mpelekh:
> @mpelekh would you be interested in joining a SIG Scalability meeting to talk through the changes?

@crenshaw-dev Yes, I’d be happy to join the SIG Scalability meeting to discuss the changes. Please let me know the time and details or if there’s anything specific I should prepare in advance.

crenshaw-dev:
Great! The event is on the [Argoproj calendar](https://calendar.google.com/calendar/u/0/embed?src=argoproj@gmail.com&pli=1), and we coordinate in [CNCF Slack](https://cloud-native.slack.com/archives/C04SURUPDL2). The next meeting is two Wednesdays from now at 8am eastern time.

No need to prepare anything really, just be prepared to answer questions about the PR. :-)

mpelekh:
> Could you open an Argo CD PR pointing to this commit so that we can run all Argo's tests?

@crenshaw-dev Sure. Here it is - https://github.com/argoproj/argo-cd/pull/20329.

crenshaw-dev:
A couple things from the contributors meeting last week:

1) we should probably make this configurable via a flag from Argo CD; the more I think about it, the more I think we should have a quick opt-out option
2) if feasible, we should have batches process on a ticker _or_ some max slice size; that'll help manage particularly high-churn spikes

mpelekh:
> Do we have a definitive answer yet for whether sync status/operation status are currently atomically updated vs. just very-quickly updated? Because if we're losing atomicity, that could be a big problem. If we're just slowing something down that used to be fast, I think that's relatively okay.

I provided the details in this comment - https://github.com/argoproj/argo-cd/pull/20329#issuecomment-2460145267

**tl;dr**
The sync and operation status are not atomically updated. They just very quickly updated.

pkg/cache/cluster.go
```diff
@@ -1240,16 +1296,65 @@ func (c *clusterCache) processEvent(event watch.EventType, un *unstructured.Unst
 		return
 	}
 
+	c.eventMetaCh <- eventMeta{event, un}
+}
+
+func (c *clusterCache) processEvents() {
+	log := c.log.WithValues("fn", "processItems")
```
andrii-korotkov-verkada, line 1303: Is "fn" a standard thing? I'd rather use a more verbose name.

andrii-korotkov-verkada, line 1303: Should we extend the lock to do this assignment, or is there nothing that can override c.log?

pkg/cache/cluster.go
```diff
@@ -1240,16 +1296,65 @@ func (c *clusterCache) processEvent(event watch.EventType, un *unstructured.Unst
...
+
+	eventMetas := make([]eventMeta, 0)
+	ticker := time.NewTicker(c.eventProcessingInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case em, ok := <-ch:
+			if !ok {
+				log.V(1).Info("Event processing channel closed, finish processing")
```
andrii-korotkov-verkada, line 1318: I think this can be debug level

pkg/cache/cluster.go
```diff
@@ -1240,16 +1296,65 @@ func (c *clusterCache) processEvent(event watch.EventType, un *unstructured.Unst
...
+	ch := c.eventMetaCh
+	c.lock.Unlock()
+
+	eventMetas := make([]eventMeta, 0)
+	ticker := time.NewTicker(c.eventProcessingInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case em, ok := <-ch:
```
andrii-korotkov-verkada, line 1316: Use a more verbose name like `eventMeta`. `em` is not a standard shortening.

pkg/cache/cluster.go
```diff
@@ -1240,16 +1296,65 @@ func (c *clusterCache) processEvent(event watch.EventType, un *unstructured.Unst
...
+			if !ok {
+				log.V(1).Info("Event processing channel closed, finish processing")
+				return
+			}
+			eventMetas = append(eventMetas, em)
+		case <-ticker.C:
+			if len(eventMetas) > 0 {
+				c.processEventsBatch(eventMetas)
+				eventMetas = eventMetas[:0]
```
andrii-korotkov-verkada, line 1325: Why assign `eventMetas[:0]` instead of nil or empty slice?

andrii-korotkov-verkada (commented):
Overall, looks great, have some minor comments.

andrii-korotkov-verkada (approved):

pkg/cache/cluster.go
mpelekh: No need to extend the lock here, as WithValues returns a new Logger instance without modifying c.log directly. The assignment to log is local and safe from concurrent modification.
https://github.com/go-logr/logr/blob/master/logr.go#L328

pkg/cache/cluster.go
mpelekh: Declaration of `eventMeta` shadows declaration at `cluster.go` - https://github.com/argoproj/gitops-engine/pull/629/files#diff-9c9e197d543705f08c9b1bc2dc404a55506cfc2935a988e6007d248257aadb1aR80.

Renamed to `evMeta`.

pkg/cache/cluster.go
mpelekh: Thank you for review @andrii-korotkov-verkada 

Setting `eventMetas` to `eventMetas[:0]` retains the original slice’s underlying array, which can help reduce memory allocations if the slice is reused frequently. Assigning `nil` would release the underlying array, and creating a new empty slice would lead to additional allocations. This approach allows us to efficiently clear the slice while keeping the capacity intact for future use.

mpelekh:
Thanks for the review @andrii-korotkov-verkada. I addressed the comments in this commit - https://github.com/argoproj/gitops-engine/pull/629/commits/7a53ecab1d0eab6197dd11221f7c0b9ed3bed738

I am going to `rebase -i --autosquash` it before merge.

pkg/cache/cluster.go
mpelekh: I completely agree with using a more verbose name. Fixed - https://github.com/argoproj/gitops-engine/commit/7a53ecab1d0eab6197dd11221f7c0b9ed3bed738

pkg/cache/cluster.go
mpelekh: It completely makes sense. Fixed - https://github.com/argoproj/gitops-engine/commit/7a53ecab1d0eab6197dd11221f7c0b9ed3bed738

rumstead (approved):

pkg/cache/settings.go
```diff
@@ -170,3 +170,17 @@ func SetRespectRBAC(respectRBAC int) UpdateSettingsFunc {
...
+
+// SetBatchEventsProcessing allows to set whether to process events in batch
+func SetBatchEventsProcessing(batchProcessing bool) UpdateSettingsFunc {
+	return func(cache *clusterCache) {
+		cache.batchEventsProcessing = batchProcessing
+	}
+}
+
+// SetEventProcessingInterval allows to set the interval for processing events
+func SetEventProcessingInterval(interval time.Duration) UpdateSettingsFunc {
```
rumstead, line 182: super nit but should we add a test asserting the default gets set without an override and vice versa?

crenshaw-dev (approved):

pkg/cache/cluster.go
```diff
@@ -185,16 +200,20 @@ func NewClusterCache(config *rest.Config, opts ...UpdateSettingsFunc) *clusterCa
 type clusterCache struct {
 	syncStatus clusterCacheSync
 
-	apisMeta      map[schema.GroupKind]*apiMeta
-	serverVersion string
-	apiResources  []kube.APIResourceInfo
+	apisMeta              map[schema.GroupKind]*apiMeta
+	batchEventsProcessing bool
+	eventMetaCh           chan eventMeta
```
crenshaw-dev, line 205: These need docstrings. I'd also rename to `batchEventsProcessingEnabled` to make clear that it's a feature flag.

pkg/cache/cluster.go
```diff
@@ -299,6 +319,29 @@ func (c *clusterCache) getEventHandlers() []OnEventHandler {
...
+}
+func (c *clusterCache) getProcessEventsHandlers() []OnProcessEventsHandler {
+	c.handlersLock.Lock()
+	defer c.handlersLock.Unlock()
+	handlers := make([]OnProcessEventsHandler, 0, len(c.processEventsHandlers))
+	for _, h := range c.processEventsHandlers {
+		handlers = append(handlers, h)
+	}
+	return handlers
+}
```
crenshaw-dev, line 343: Why not have the caller just use the map directly?

pkg/cache/cluster.go
```diff
@@ -926,6 +983,14 @@ func (c *clusterCache) sync() error {
 	return nil
 }
 
+// invalidateEventMeta closes the eventMeta channel if it is open
```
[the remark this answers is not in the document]
crenshaw-dev: @mpelekh can you update the go doc?

PR #634, opened 2024-11-07
fix: Ability to disable Server Side Apply on individual resource level

pasha-codefresh:
Closes: https://github.com/argoproj/argo-cd/issues/20306

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -948,6 +948,23 @@ func (sc *syncContext) ensureCRDReady(name string) error {
 	})
 }
 
+func (sc *syncContext) shouldUseServerSideApply(targetObj *unstructured.Unstructured) bool {
+	// if it is a dry run, disable server side apply, as the goal is to validate only the
+	// yaml correctness of the rendered manifests.
+	// running dry-run in server mode breaks the auto create namespace feature
+	// https://github.com/argoproj/argo-cd/issues/13874
+	if sc.dryRun {
+		return false
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/sync/sync_context.go
```diff
@@ -948,6 +948,23 @@ func (sc *syncContext) ensureCRDReady(name string) error {
 	})
 }
 
+func (sc *syncContext) shouldUseServerSideApply(targetObj *unstructured.Unstructured) bool {
+	// if it is a dry run, disable server side apply, as the goal is to validate only the
+	// yaml correctness of the rendered manifests.
+	// running dry-run in server mode breaks the auto create namespace feature
+	// https://github.com/argoproj/argo-cd/issues/13874
+	if sc.dryRun {
```
leoluz, line 956: ServerSide diff uses ServerSide Apply in dryrun mode. We need to make sure that the ServerSide diff code path doesn't call this function. Another option is adding a new param to the function to determine the dryrun behaviour.

leoluz (commented):
Can you just please confirm if the ServerSide Diff code path is not affected by this change?

pkg/sync/sync_context.go
pasha-codefresh: I didnt change this part, it was already here. before we had 
!dryRun && (sc.serverSideApply || resourceutil.HasAnnotationOption(t.targetObj, common.AnnotationSyncOptions, common.SyncOptionServerSideApply))

so if dryRun is false, it will use client side apply

leoluz (approved):
LGTM

PR #637, opened 2024-11-17
chore(deps): Upgrade structured-merge-diff from v4.4.1 to v4.4.3

andrii-korotkov-verkada:
Adding updates from the last year, let's see if that fixed some bugs


thecosmicfrog (approved):
Approved based on my bug report here: https://github.com/argoproj/argo-cd/issues/20792

jenting (approved):

crenshaw-dev (approved):
Doesn't fix the bug, so won't cherry-pick. But the version bump makes sense to go into 2.14.

PR #640, opened 2024-11-18
fix: Server side diff now works correctly with some fields removal

andrii-korotkov-verkada:
Helps with https://github.com/argoproj/argo-cd/issues/20792

Removed and modified sets may only contain the fields that changed, not including key fields like "name". This can cause merge to fail, since it expects those fields to be present if they are present in the predicted live. Fortunately, structured merge diff folks have added an option to extracting items to add key fields, and we are going to use that.

Failure of the new test before the fix
```
            	Error:      	Received unexpected error:
            	            	error removing non config mutations for resource Deployment/nginx-deployment: error reverting webhook removed fields in predicted live resource: .spec.template.spec.containers: element 0: associative list with keys has an element that omits key field "name" (and doesn't have default value)
            	Test:       	TestServerSideDiff/will_test_removing_some_field_with_undoing_changes_done_by_webhook
```

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -258,7 +258,13 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
 	}
 
 	if comparison.Modified != nil && !comparison.Modified.Empty() {
-		liveModValues := typedLive.ExtractItems(comparison.Modified)
+		comparisonModifiedForMerge := comparison.Modified
+		typedPredictedLiveFieldSet, err := typedPredictedLive.ToFieldSet()
+		if err != nil {
+			return nil, fmt.Errorf("error reverting webhook modified fields in predicted live resource: %s", err)
+		}
+		appendKeyFieldsToRhs(typedPredictedLiveFieldSet, comparisonModifiedForMerge)
+		liveModValues := typedLive.ExtractItems(comparisonModifiedForMerge)
 		// revert modified fields not owned by any manager
 		typedPredictedLive, err = typedPredictedLive.Merge(liveModValues)
 		if err != nil {
@@ -267,7 +273,13 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
 	}
 
 	if comparison.Removed != nil && !comparison.Removed.Empty() {
-		liveRmValues := typedLive.ExtractItems(comparison.Removed)
+		comparisonRemovedForMerge := comparison.Removed
+		typedPredictedLiveFieldSet, err := typedPredictedLive.ToFieldSet()
+		if err != nil {
+			return nil, fmt.Errorf("error reverting webhook removed fields in predicted live resource: %s", err)
+		}
+		appendKeyFieldsToRhs(typedPredictedLiveFieldSet, comparisonRemovedForMerge)
+		liveRmValues := typedLive.ExtractItems(comparisonRemovedForMerge)
 		// revert removed fields not owned by any manager
 		typedPredictedLive, err = typedPredictedLive.Merge(liveRmValues)
 		if err != nil {
@@ -281,6 +293,31 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
 		return nil, fmt.Errorf("error converting live typedValue: expected map got %T", plu)
 	}
 	return &unstructured.Unstructured{Object: pl}, nil
+}
+
+// appendKeyFieldsToRhs appends the key fields like "name" to the rhs elements which have some entries keyed by those fields.
+// The fields are only appended if they are present in lhs as well, otherwise assuming they have a default value not needed
+// to be specified explicitly.
+// This is needed to merge properly, see https://github.com/argoproj/argo-cd/issues/20792.
+func appendKeyFieldsToRhs(lhs *fieldpath.Set, rhs *fieldpath.Set) {
+	keyFieldPaths := []fieldpath.Path{}
+	rhs.Iterate(func(path fieldpath.Path) {
+		for i := 0; i < len(path); i++ {
+			if path[i].Key != nil {
+				for _, keyField := range *path[i].Key {
+					pathPartCopy := make([]fieldpath.PathElement, i+1)
+					copy(pathPartCopy, path[:i+1])
+					newPath := append(pathPartCopy, fieldpath.PathElement{FieldName: &keyField.Name})
+					keyFieldPaths = append(keyFieldPaths, newPath)
+				}
+			}
+		}
+	})
+	for _, path := range keyFieldPaths {
+		if !rhs.Has(path) && lhs.Has(path) {
+			rhs.Insert(path)
+		}
+	}
 }
 
 func jsonStrToUnstructured(jsonString string) (*unstructured.Unstructured, error) {
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -258,13 +258,7 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
 	}
 
 	if comparison.Modified != nil && !comparison.Modified.Empty() {
-		comparisonModifiedForMerge := comparison.Modified
-		typedPredictedLiveFieldSet, err := typedPredictedLive.ToFieldSet()
-		if err != nil {
-			return nil, fmt.Errorf("error reverting webhook modified fields in predicted live resource: %s", err)
-		}
-		appendKeyFieldsToRhs(typedPredictedLiveFieldSet, comparisonModifiedForMerge)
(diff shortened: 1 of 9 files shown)


pkg/diff/diff_test.go
```diff
@@ -871,7 +871,7 @@ func TestStructuredMergeDiff(t *testing.T) {
 		// then
 		require.NoError(t, err)
 		assert.NotNil(t, result)
-		assert.False(t, result.Modified)
+		assert.True(t, result.Modified)
```
crenshaw-dev, line 874: Why does this result in a diff when it didn't before?

pkg/diff/diff.go
```diff
@@ -283,6 +285,28 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
 	return &unstructured.Unstructured{Object: pl}, nil
 }
 
+// Appends the key fields like "name" to the set elements which have some entries keyed by those fields.
+// This is needed to merge properly, see https://github.com/argoproj/argo-cd/issues/20792.
+func appendKeyFields(set *fieldpath.Set) {
```
crenshaw-dev, line 290:
```suggestion
// appendKeyFields appends the key fields like "name" to the set elements which have some entries keyed by those fields.
// This is needed to merge properly, see https://github.com/argoproj/argo-cd/issues/20792.
func appendKeyFields(set *fieldpath.Set) {
```

pkg/diff/diff.go
```diff
@@ -283,6 +285,28 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
...
+	// Map is to deduplicate
```
crenshaw-dev, line 291: Can you expand the comment to explain why de-duplication is necessary? What goes wrong if we don't de-dupe?

pkg/diff/diff.go
crenshaw-dev, line 290: Is there a good way to unit test this function? I understand it's covered by higher-level function tests, but I think some tests specific to this function would really clarify how it works.

pkg/diff/diff.go
```diff
@@ -258,6 +258,7 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
 	}
 
 	if comparison.Modified != nil && !comparison.Modified.Empty() {
```
crenshaw-dev, line 260: Is this bug also potentially an issue for added fields? If not, why not? Maybe should add a comment to explain.

crenshaw-dev (changes requested):
Can you open an Argo CD PR to test this? Is there a way to test the higher-level bug in Argo CD so we have coverage of both the gitops-engine behavior and of the user-facing bug?

pkg/diff/diff_test.go
andrii-korotkov-verkada: I've updated the deployment to actually have some changes, to better replicate the behavior from the bug report.

andrii-korotkov-verkada:
> Can you open an Argo CD PR to test this? Is there a way to test the higher-level bug in Argo CD so we have coverage of both the gitops-engine behavior and of the user-facing bug?

I'll do the testing.

pkg/diff/diff.go
andrii-korotkov-verkada: Updating

pkg/diff/diff.go
andrii-korotkov-verkada: I ended up replacing with a list, since a small optimization of reducing a number of calls to Has below might be overshadowed by the extra cost for a string conversion.

pkg/diff/diff.go
andrii-korotkov-verkada: I'll look into this

pkg/diff/diff.go
andrii-korotkov-verkada: We don't call merge, so maybe not, but I'll check.

pkg/diff/diff.go
andrii-korotkov-verkada: Glad I did a test - it caught a subtle bug.

pkg/diff/diff.go
```diff
@@ -283,6 +285,29 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
...
+// This is needed to merge properly, see https://github.com/argoproj/argo-cd/issues/20792.
+func appendKeyFields(set *fieldpath.Set) {
+	keyFieldPaths := []fieldpath.Path{}
+	set.Iterate(func(path fieldpath.Path) {
+		for i := 0; i < len(path); i++ {
+			if path[i].Key != nil {
+				for _, keyField := range *path[i].Key {
+					pathPartCopy := make([]fieldpath.PathElement, i+1)
+					copy(pathPartCopy, path[:i+1])
+					newPath := append(pathPartCopy, fieldpath.PathElement{FieldName: &keyField.Name})
```
leoluz, line 298: You are assuming that the field `name` is always the key in lists which can be true or false depending on the resource. We can't assume schema structure from this library and must rely on the provided `GvkParser`. The problem is likely cause by the fact that the `GvkParser` isn't populated with the necessary default values and list keys for the specific schema. It is strange, however that the error is being reported from a `Deployment` resource as this should contain the necessary schema info IRRC. 
Please investigate more and let me know your findings.

leoluz (changes requested):
Please check my comment.

pkg/diff/diff.go
leoluz: @andrii-korotkov-verkada I understand. My point is, how you determine what is the key that should be considered when merging 2 hypothetical lists?

andrii-korotkov-verkada: All of the keys should be considered as far as I understand. Let me take a further look to see if we can derive from schema.

andrii-korotkov-verkada: I also don't know if key fields are supposed to be returned by the diff, e.g. since they aren't necessarily removed. It's just we use those fields in a special way to do a merge, which has a different requirements that diff computation can have.

pkg/diff/testdata/smd-deploy-live.yaml
```diff
@@ -120,7 +125,10 @@ spec:
           ports:
             - containerPort: 80
               protocol: TCP
-          resources: {}
+          resources:
+            requests:
+              memory: 512Mi
+              cpu: 500m
```
crenshaw-dev, line 131: Rather than change the existing test, should we create a new test and leave the old one in place?

pkg/diff/diff.go
andrii-korotkov-verkada: For example, we don't want key fields in added items, since we remove them in this logic. So not sure default behavior of the libs should be changed and it's on us to adjust, which I'm trying to do with this PR.

pkg/diff/testdata/smd-deploy-live.yaml
```diff
@@ -120,7 +125,10 @@ spec:
           ports:
             - containerPort: 80
               protocol: TCP
```
crenshaw-dev, line 127: In the case of both `ports` and `containers`, bot the live and the config yaml contain only one list item. I think we should probably do a test involving multiple list items, ideally with differing orders, to demonstrate that the key matching works correctly.

crenshaw-dev (changes requested):

pkg/diff/testdata/smd-deploy-live.yaml
andrii-korotkov-verkada: Alright, I'll create a 2nd deployment manifests files.

pkg/diff/testdata/smd-deploy-live.yaml
andrii-korotkov-verkada: Okay

andrii-korotkov-verkada:
@crenshaw-dev, I've addressed your comments.

Code Analysis complains that configs are not perfect, but that's okay since they aren't deployed anywhere.

andrii-korotkov-verkada:
Per discussion with @leoluz, I'd spend more time to understand why the Structured Merge Diff library's Merge doesn't derive key fields from paths, since the info is available. I may submit a ticket to their community with a request to update the library.

andrii-korotkov-verkada:
I've filed https://github.com/kubernetes-sigs/structured-merge-diff/issues/273

leoluz:
Structured-merge-diff PR with the fix is now merged:
https://github.com/kubernetes-sigs/structured-merge-diff/pull/275

sonar-project.properties
```diff
@@ -0,0 +1,2 @@
+# Exclude test manifests from analysis
+sonar.kubernetes.exclusions=pkg/diff/testdata/**
```
leoluz, line 2: This doesn't seem to be working

leoluz (approved):
LGTM

PR #644, opened 2024-11-27
feat: add health check support for ArgoCD Application

ksdpmx:
When using a centralized ArgoCD instance in the control plane, it may sometimes be necessary to configure an ArgoCD Application as a hook, e.g. running an application as a post-sync hook, and specifying a destination cluster to execute jobs in a data plane cluster.

But right now the health check for this CRD is not supported in ArgoCD (or [Helm either](https://github.com/helm/helm/blob/main/pkg/kube/client.go#L750-L753)).. which makes it hard for the controller to determine whether to clean up the hook or not when using the delete policy HookSucceeded.

### DIFF
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -120,6 +120,8 @@ func GetHealthCheckFunc(gvk schema.GroupVersionKind) func(obj *unstructured.Unst
 		switch gvk.Kind {
 		case "Workflow":
 			return getArgoWorkflowHealth
+		case "Application":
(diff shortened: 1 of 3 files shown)


crenshaw-dev:
Closing since this should be implemented as a Lua script on the argo-cd repo. But I'm supportive of the change! We'll just need to be very very careful to get the check logic in good shape and communicate via release notes that people will need to disable it if they don't want the check.

PR #646, opened 2024-12-04
fix: stuck hook issue when a Job resource has a ttlSecondsAfterFinished field set

dejanzele:
This is a proof-of-concept PR which would fix https://github.com/argoproj/argo-cd/issues/21055

More info on the issue can be found in the linked GitHub issue.

~~Idea is to add a finalizer to Job resources which have `ttlSecondsAfterFinished` set and remove it after ArgoCD detects the hook completed.~~

~~A simpler approach would be to unset the `ttlSecondsAfterFinished` but that would cause drift from defined vs actual state.~~

The proposed solution is to attach a finalizer on all hook tasks and remove it after the argocd acknowledges the hook task is completed in the sync phase.

The same scenario which is described in the linked GitHub issue passes in this PR.

I welcome any feedback, as I think a lot of people on the community would like this to be fixed, and I'd be more than happy to adopt based on best direction for this issue.


pkg/sync/sync_context.go
```diff
@@ -830,6 +850,83 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
...
+
+		job, err := sc.clientset.BatchV1().Jobs(task.targetObj.GetNamespace()).Get(context.TODO(), task.targetObj.GetName(), metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		// Skip postprocessing if the Job does not have a ttlSecondsAfterFinished set.
+		if job.Spec.TTLSecondsAfterFinished == nil {
+			return nil
+		}
```
crenshaw-dev, line 888: What would be the downside(s) to always using a finalizer instead of only using one when this field is set?

pkg/sync/sync_context.go
```diff
@@ -830,6 +850,83 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
...
+		// Currently, we only process hook tasks where the target object is a Job.
+		if task.targetObj.GetKind() == "Job" {
+			sc.processJobHookTask(task)
+		}
+	}
+}
+
+// processJobHookTask processes a hook task where the target object is a Job and has defined ttlSecondsAfterFinished.
+// This addresses the issue where a Job with a ttlSecondsAfterFinished set to a low value gets deleted fast and the hook phase gets stuck.
```
crenshaw-dev, line 872: Should the finalizer feature be limited to just Jobs? I understand that Argo Workflows can also exhibit the same behavior. I could imagine any resource having the same issue if some process deletes the resource before Argo CD has a chance to observe the final resource state.

pkg/sync/sync_context.go
dejanzele: That is a good question, the logic would be simpler and more generic if all hooks have a finalizer, cleanup would also be simpler.

pkg/sync/sync_context.go
dejanzele: I thought about that, and yes, that is true.
I focused only on the immediate issue as a PoC, I am also in favour of a generic solution for the scenario where some external process deletes a resource during hook phase.

pkg/sync/sync_context.go
```diff
@@ -577,6 +587,29 @@ func (sc *syncContext) filterOutOfSyncTasks(tasks syncTasks) syncTasks {
...
+	var mutated bool
+	for i, finalizer := range finalizers {
+		if finalizer == hook.HookFinalizer {
+			finalizers = append(finalizers[:i], finalizers[i+1:]...)
+			mutated = true
+			break
+		}
+	}
+	if mutated {
+		task.targetObj.SetFinalizers(finalizers)
```
dejanzele, line 604: Should I set it at all for the `targetObj`?

dejanzele:
@crenshaw-dev I have updated the PR to make the solution more generic.

This use case should be also covered by an integration test, but I guess that test would live in the argocd repository?

crenshaw-dev:
@dejanzele yep! You can open a PR on the argo-cd repo temporarily replacing gitops-engine in go.mod with your fork and revision.

pkg/sync/sync_context.go
crenshaw-dev: At a glance, it's not clear to me.... would you mind investigating/documenting that `liveObj` and `targetObj` are on the `syncTask` struct?

pkg/sync/sync_context.go
```diff
@@ -577,6 +586,56 @@ func (sc *syncContext) filterOutOfSyncTasks(tasks syncTasks) syncTasks {
...
+		task.liveObj.SetFinalizers(finalizers)
+		// The cached live object may be stale in the controller cache, and the actual object may have been updated in the meantime,
+		// and Kubernetes API will return a conflict error on the Update call.
+		// In that case, we need to get the latest version of the object and retry the update.
+		return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			updateErr := sc.updateResource(task)
+			if apierr.IsConflict(updateErr) {
```
dejanzele, line 610: Conflicts happen quite often and without retries the E2E tests were running very flaky.

agaudreault (approved):
Tested and updated with latest master. LGTM!

PR #653, opened 2024-12-13
feat: New sync flags for replace and force

andrii-korotkov-verkada:
Gitops-engine part of https://github.com/argoproj/argo-cd/issues/21427
    
Add new way of specifying options and include options always and never. Also, include a way to specify options per group/kind.

### DIFF
--- a/pkg/sync/common/types.go
+++ b/pkg/sync/common/types.go
@@ -28,8 +28,16 @@ const (
 	SyncOptionPruneLast = "PruneLast=true"
 	// Sync option that enables use of replace or create command instead of apply
 	SyncOptionReplace = "Replace=true"
+	// Sync option that enables use of replace or create command instead of apply
+	SyncOptionReplaceAlways = "Replace=always"
+	// Sync option that disables use of replace or create command instead of apply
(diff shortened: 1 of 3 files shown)


leoluz (changes requested):
I am not sure if this is the appropriate solution to the problem. I am mainly concerned with confusion that this would bring to users when there is no clear indication of how Argo CD is applying resources. As discussed in the contrib meeting, we need to better understand the edge cases that this would bring (e.g. how to manage conflicts with UI configuration during syncs)

andrii-korotkov-verkada:
> I am not sure if this is the appropriate solution to the problem. I am mainly concerned with confusion that this would bring to users when there is no clear indication of how Argo CD is applying resources. As discussed in the contrib meeting, we need to better understand the edge cases that this would bring (e.g. how to manage conflicts with UI configuration during syncs)

The alternative seems worse due to autosync just not working when there's a mix of resources that need to be synced with replace always and replace never.

With my approach I don't see any nasty edge cases - we have one sync and "Replace (if allowed)" checkbox. Just resources which should be replaced are replaced and resources which aren't allowed to be replaced are synced normally. Do you have any specific edge cases in mind?

PR #661, opened 2025-01-31
chore: New KubeApplier for server side diff dry run

andrii-korotkov-verkada:
Helps with: https://github.com/argoproj/argo-cd/issues/21488

Implement a new kube applier that can be derived from resource opts (so we don't copy management code and worry about overlaps). It would be used by ArgoCD when passing the kube appplier. The difference is whether to capture stderr from the kube command. For server side diff dry run we don't.

### DIFF
--- a/pkg/utils/kube/resource_ops.go
+++ b/pkg/utils/kube/resource_ops.go
@@ -43,6 +43,7 @@ type ResourceOperations interface {
 	ReplaceResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force bool) (string, error)
 	CreateResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, validate bool) (string, error)
 	UpdateResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy) (*unstructured.Unstructured, error)
+	GetServerSideDiffDryRunApplier() diff.KubeApplier
 }
 
 type kubectlResourceOperations struct {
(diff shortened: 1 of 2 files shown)


pkg/utils/kube/resource_ops.go
```diff
@@ -43,6 +43,7 @@ type ResourceOperations interface {
 	ReplaceResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force bool) (string, error)
 	CreateResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, validate bool) (string, error)
 	UpdateResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy) (*unstructured.Unstructured, error)
+	GetServerSideDiffDryRunApplier() diff.KubeApplier
```
leoluz, line 46: I don't think that we should add a new operation to this interface.

pkg/utils/kube/resource_ops.go
```diff
@@ -244,7 +262,11 @@ func (k *kubectlResourceOperations) UpdateResource(ctx context.Context, obj *uns
 
 // ApplyResource performs an apply of a unstructured resource
 func (k *kubectlResourceOperations) ApplyResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string, serverSideDiff bool) (string, error) {
-	span := k.tracer.StartSpan("ApplyResource")
+	return k.applyResourceImpl(ctx, obj, dryRunStrategy, force, validate, serverSideApply, manager, serverSideDiff, true)
+}
+
+func (k *kubectlResourceOperations) applyResourceImpl(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string, serverSideDiff, appendStderr bool) (string, error) {
```
leoluz, line 268: The point about having a new interface is to avoid adding new parameters in the existing functions. `appendStderr bool` shouldn't be added here.

leoluz (changes requested):
Please check my comments.

pkg/utils/kube/resource_ops.go
andrii-korotkov-verkada: It's a helper function, that the original function uses. If we don't do that, we'd get copy-paste.

pkg/utils/kube/resource_ops.go
andrii-korotkov-verkada: What's the issue with that?

andrii-korotkov-verkada: Otherwise I need to init a whole new object for kubectl commands and it'd be a lot of extra work and duplication.

leoluz: `GetServerSideDiffDryRunApplier` <- This operation does not belong to this interface. Interfaces should be as lean and concise as possible. This is mainly a shortcut and it is better if we remove it.

leoluz (changes requested):

pkg/utils/kube/resource_ops.go
andrii-korotkov-verkada: `getResourceOperations` in ArgoCD returns `kube.ResourceOperations`, thus I don't see how to avoid this in the interface without initializing a whole new kubectl cmd variant, which seems like extra overhead.

leoluz:
closing it in favour of https://github.com/argoproj/gitops-engine/pull/662

PR #662, opened 2025-02-03
fix: New kube applier for server side diff dry run with refactoring

andrii-korotkov-verkada:
Part of a fix for: https://github.com/argoproj/argo-cd/issues/21488

Separate logic to handle server side diff dry run applies.

### DIFF
--- a/pkg/utils/kube/resource_ops.go
+++ b/pkg/utils/kube/resource_ops.go
@@ -39,7 +39,7 @@ import (
 
 // ResourceOperations provides methods to manage k8s resources
 type ResourceOperations interface {
-	ApplyResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string, serverSideDiff bool) (string, error)
+	ApplyResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string) (string, error)
 	ReplaceResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force bool) (string, error)
 	CreateResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, validate bool) (string, error)
 	UpdateResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy) (*unstructured.Unstructured, error)
@@ -298,7 +298,7 @@ func (k *kubectlResourceOperations) UpdateResource(ctx context.Context, obj *uns
 }
 
 // ApplyResource performs an apply of a unstructured resource
-func (k *kubectlServerSideDiffDryRunApplier) ApplyResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string, serverSideDiff bool) (string, error) {
+func (k *kubectlServerSideDiffDryRunApplier) ApplyResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string) (string, error) {
 	span := k.tracer.StartSpan("ApplyResource")
 	span.SetBaggageItem("kind", obj.GetKind())
 	span.SetBaggageItem("name", obj.GetName())
@@ -306,8 +306,7 @@ func (k *kubectlServerSideDiffDryRunApplier) ApplyResource(ctx context.Context, 
 	k.log.WithValues(
 		"dry-run", [...]string{"none", "client", "server"}[dryRunStrategy],
 		"manager", manager,
-		"serverSideApply", serverSideApply,
-		"serverSideDiff", serverSideDiff).Info(fmt.Sprintf("Applying resource %s/%s in cluster: %s, namespace: %s", obj.GetKind(), obj.GetName(), k.config.Host, obj.GetNamespace()))
+		"serverSideApply", serverSideApply).Info(fmt.Sprintf("Applying resource %s/%s in cluster: %s, namespace: %s", obj.GetKind(), obj.GetName(), k.config.Host, obj.GetNamespace()))
 
 	return k.runResourceCommand(obj, func(ioStreams genericclioptions.IOStreams, fileName string) error {
 		cleanup, err := processKubectlRun(k.onKubectlRun, "apply")
@@ -325,7 +324,7 @@ func (k *kubectlServerSideDiffDryRunApplier) ApplyResource(ctx context.Context, 
 }
 
 // ApplyResource performs an apply of a unstructured resource
-func (k *kubectlResourceOperations) ApplyResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string, serverSideDiff bool) (string, error) {
+func (k *kubectlResourceOperations) ApplyResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string) (string, error) {
 	span := k.tracer.StartSpan("ApplyResource")
 	span.SetBaggageItem("kind", obj.GetKind())
 	span.SetBaggageItem("name", obj.GetName())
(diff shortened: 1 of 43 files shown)


pkg/utils/kube/resource_ops.go
```diff
@@ -125,6 +139,47 @@ func (k *kubectlResourceOperations) runResourceCommand(ctx context.Context, obj
...
+		return "", fmt.Errorf("Failed to write manifest: %v", err)
+	}
+	if err = manifestFile.Close(); err != nil {
+		return "", fmt.Errorf("Failed to close manifest: %v", err)
+	}
+
+	err = maybeLogManifest(manifestBytes, k.log)
+	if err != nil {
+		return "", err
+	}
```
leoluz, line 162: This code is a duplication of what is defined in the kubectlResourceOperations.runResourceCommand() method. Please extract this code in a private function and invoke it by both `runResourceCommand` methods.

pkg/utils/kube/resource_ops.go
```diff
@@ -125,6 +139,47 @@ func (k *kubectlResourceOperations) runResourceCommand(ctx context.Context, obj
...
+		In:     &bytes.Buffer{},
+		Out:    &bytes.Buffer{},
+		ErrOut: &bytes.Buffer{},
+	}
+	err = executor(ioStreams, manifestFile.Name())
+	if err != nil {
+		return "", errors.New(cleanKubectlOutput(err.Error()))
+	}
+	if buf := strings.TrimSpace(ioStreams.ErrOut.(*bytes.Buffer).String()); len(buf) > 0 {
+		k.log.Info("Server-side dry run apply had non-empty stderr: %s", buf)
```
leoluz, line 175: I think that if we get something in stderr, we should log and return the ErrOut as an error

pkg/utils/kube/resource_ops.go
```diff
@@ -243,7 +298,33 @@ func (k *kubectlResourceOperations) UpdateResource(ctx context.Context, obj *uns
...
-func (k *kubectlResourceOperations) ApplyResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string, serverSideDiff bool) (string, error) {
+func (k *kubectlServerSideDiffDryRunApplier) ApplyResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string) (string, error) {
+	span := k.tracer.StartSpan("ApplyResource")
+	span.SetBaggageItem("kind", obj.GetKind())
+	span.SetBaggageItem("name", obj.GetName())
+	defer span.Finish()
+	k.log.WithValues(
+		"dry-run", [...]string{"none", "client", "server"}[dryRunStrategy],
+		"manager", manager,
+		"serverSideApply", serverSideApply).Info(fmt.Sprintf("Applying resource %s/%s in cluster: %s, namespace: %s", obj.GetKind(), obj.GetName(), k.config.Host, obj.GetNamespace()))
```
leoluz, line 309: We should change the log message as now we know that this is not a real apply. Something like: `Running ServerSide Diff: Applying resource %s/%s in cluster: %s, namespace: %s`

pkg/utils/kube/resource_ops.go
```diff
@@ -316,32 +413,46 @@ func (k *kubectlResourceOperations) newApplyOptions(ioStreams genericclioptions.
 				return nil, err
 			}
 		case cmdutil.DryRunServer:
```
leoluz, line 415: We don't need this switch/case anymore in this case as ServerSideDiff must always be set with `DryrunServer` strategy. Maybe we can add a validation here and check if `o.DryRunStrategy != cmdutil.DryRunServer` and return error in this case.

pkg/utils/kube/resource_ops.go
```diff
@@ -54,9 +54,39 @@ type kubectlResourceOperations struct {
 	openAPISchema openapi.Resources
 }
 
-type commandExecutor func(f cmdutil.Factory, ioStreams genericclioptions.IOStreams, fileName string) error
+type kubectlServerSideDiffDryRunApplier struct {
```
leoluz, line 57: Can you please add a small doc here explaining that this is a KubeApplier implementation specifically for ServerSide Diff?

leoluz (changes requested):
PR looking good!
Please check my comments.
Thank you.

pkg/utils/kube/resource_ops.go
andrii-korotkov-verkada: Maybe as a warning, since the command itself didn't fail.

pkg/utils/kube/resource_ops.go
andrii-korotkov-verkada: Okay, will try.

pkg/utils/kube/resource_ops.go
andrii-korotkov-verkada: Makes sense

pkg/utils/kube/resource_ops.go
andrii-korotkov-verkada: Ah, logr doesn't support warning

andrii-korotkov-verkada: I'd rather keep it as info, errors can potentially result in spam. Looks like level 0 (info) is used for both info and warn, e.g. https://pkg.go.dev/github.com/go-logr/logr#ToSlogHandler.

leoluz: My point was less about the log level and more about returning an error. Currently, even if the kubectl apply returns an error in stderr, we just log and continue with the function execution. It seems to me that instead we should return the error and stop the execution.

pkg/utils/kube/resource_ops.go
leoluz: Nit: Mainly for clarity, would you mind adding the name of the interface (`KubeApplier`) in the comment as well?

pkg/utils/kube/resource_ops.go
andrii-korotkov-verkada: One example of what can be there is
```
Warning: resource configmaps/<placeholder> is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by  apply.  apply should only be used on resources created declaratively by either  create --save-config or  apply. The missing annotation will be patched automatically.'
```
This seems like something that would be fixed automatically, so we may just proceed.

pkg/utils/kube/resource_ops.go
andrii-korotkov-verkada: Sure

pkg/utils/kube/resource_ops.go
andrii-korotkov-verkada: Will update as discussed offline.

pkg/utils/kube/resource_ops.go
```diff
@@ -125,6 +149,40 @@ func (k *kubectlResourceOperations) runResourceCommand(ctx context.Context, obj
...
+	stdout := stdoutBuf.String()
+	stderr := stderrBuf.String()
+
+	if stderr != "" && stdout == "" {
+		err := fmt.Errorf("Server-side dry run apply had non-empty stderr: %s", stderr)
+		k.log.Error(err, "server-side diff")
+		return "", err
+	}
+	if stderr != "" {
+		k.log.Info("Warning: Server-side dry run apply had non-empty stderr: %s", stderr)
```
crenshaw-dev, line 181: Should we just go ahead and log this as a warning?

pkg/utils/kube/resource_ops.go
```diff
@@ -243,7 +301,33 @@ func (k *kubectlResourceOperations) UpdateResource(ctx context.Context, obj *uns
...
-func (k *kubectlResourceOperations) ApplyResource(ctx context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string, serverSideDiff bool) (string, error) {
+func (k *kubectlServerSideDiffDryRunApplier) ApplyResource(_ context.Context, obj *unstructured.Unstructured, dryRunStrategy cmdutil.DryRunStrategy, force, validate, serverSideApply bool, manager string) (string, error) {
+	span := k.tracer.StartSpan("ApplyResource")
+	span.SetBaggageItem("kind", obj.GetKind())
+	span.SetBaggageItem("name", obj.GetName())
+	defer span.Finish()
+	k.log.WithValues(
+		"dry-run", [...]string{"none", "client", "server"}[dryRunStrategy],
+		"manager", manager,
+		"serverSideApply", serverSideApply).Info(fmt.Sprintf("Running server-side diff. Dry run applying resource %s/%s in cluster: %s, namespace: %s", obj.GetKind(), obj.GetName(), k.config.Host, obj.GetNamespace()))
```
crenshaw-dev, line 312: Maybe debug instead of info?

pkg/utils/kube/resource_ops.go
andrii-korotkov-verkada: Logger doesn't have a concept of warning, only info, error and different log levels.

pkg/utils/kube/resource_ops.go
andrii-korotkov-verkada: Yeah, dry-run probably okay to make debug

crenshaw-dev: Does make me wonder why they're hitting this code so much. Probably suspect cache behavior. Either way, debug is probably fine.

andrii-korotkov-verkada: https://github.com/argoproj/gitops-engine/pull/705

PR #664, opened 2025-02-06
WIP: Update the sync logic so that namespace are created first and the res…

Nheijmink19:
…ources in the namespace afterwards

I changed the sync logic so that when namespace creation is required, the namespace is created first. This solves the issue where argocd fails on the validation of resource creation because of the lack of permissions when creation rights are handled by rolebindings in the namespace rather than clusterrolebindings.

I created this example solution for a ticket raised on the argo-cd project: https://github.com/argoproj/argo-cd/issues/21788

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -452,6 +452,13 @@ func (sc *syncContext) Sync() {
 				}
 			}
 		}
+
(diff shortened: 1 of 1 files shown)


Nheijmink19:
In our discussion we are heading into a different direction. Closing this pull request for that reason

PR #674, opened 2025-02-07
chore: enable early-return, indent-error-flow and unnecessary-stmt from revive

mmorel-35:


pkg/health/health_daemonset.go
```diff
@@ -27,31 +27,30 @@ func getDaemonSetHealth(obj *unstructured.Unstructured) (*HealthStatus, error) {
...
-			}, nil
-		}
-		if daemon.Status.NumberAvailable < daemon.Status.DesiredNumberScheduled {
-			return &HealthStatus{
-				Status:  HealthStatusProgressing,
-				Message: fmt.Sprintf("Waiting for daemon set %q rollout to finish: %d of %d updated pods are available...", daemon.Name, daemon.Status.NumberAvailable, daemon.Status.DesiredNumberScheduled),
-			}, nil
-		}
-	} else {
+	if !(daemon.Generation <= daemon.Status.ObservedGeneration) {
```
crenshaw-dev, line 30:
```suggestion
	if daemon.Generation > daemon.Status.ObservedGeneration {
```

would this still be correct?

crenshaw-dev (approved):

PR #683, opened 2025-02-12
fix: Server-side diff shows incorrect diffs for list related changes

pjiang-dev:
fixes https://github.com/argoproj/argo-cd/issues/21817

The current Difference function does not remove every field and its children from the set. This causes unwanted the behavior of nested fields not being removed resulting in an incorrect server side diff. Using the RecursiveDifference function instead will do this:
"this removes every field **and its children** from s that is contained in s2."

Added additional unit test to cover the nested fields scenario.
Before fix:
Screenshot 2025-02-12 at 10 55 32 AM

After fix:
Screenshot 2025-02-12 at 10 55 10 AM

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -13,7 +13,7 @@ import (
 	"fmt"
 	"reflect"
 
-	jsonpatch "github.com/evanphx/json-patch"
+	jsonpatch "github.com/evanphx/json-patch/v5"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.22.0
 
 require (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
-	github.com/evanphx/json-patch v4.12.0+incompatible
+	github.com/evanphx/json-patch/v5 v5.9.11
 	github.com/go-logr/logr v1.4.2
 	github.com/golang/mock v1.6.0
 	github.com/google/gnostic-models v0.6.8
--- a/go.sum
+++ b/go.sum
@@ -33,8 +33,8 @@ github.com/emicklei/go-restful/v3 v3.11.0 h1:rAQeMHw1c7zTmncogyy8VvRZwtkmkZ4FxER
 github.com/emicklei/go-restful/v3 v3.11.0/go.mod h1:6n3XBCmQQb25CM2LCACGz8ukIrRry+4bhvbpWn3mrbc=
 github.com/envoyproxy/go-control-plane v0.9.1-0.20191026205805-5f8ba28d4473/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
 github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
-github.com/evanphx/json-patch v4.12.0+incompatible h1:4onqiflcdA9EOZ4RxV643DvftH5pOlLGNtQ5lPWQu84=
-github.com/evanphx/json-patch v4.12.0+incompatible/go.mod h1:50XU6AFN0ol/bzJsmQLiYLvXMP4fmwYFNcr97nuDLSk=
+github.com/evanphx/json-patch/v5 v5.9.11 h1:/8HVnzMq13/3x9TPvjG08wUGqBTmZBsCWzjTM0wiaDU=
+github.com/evanphx/json-patch/v5 v5.9.11/go.mod h1:3j+LviiESTElxA4p3EMKAB9HXj3/XEtnUf6OZxqIQTM=
 github.com/exponent-io/jsonpath v0.0.0-20151013193312-d6023ce2651d h1:105gxyaGwCFad8crR9dcMQWvV9Hvulu6hwUh4tWPJnM=
 github.com/exponent-io/jsonpath v0.0.0-20151013193312-d6023ce2651d/go.mod h1:ZZMPRZwes7CROmyNKgQzC3XPs6L/G2EJLHddWejkmf4=
 github.com/fatih/camelcase v1.0.0 h1:hxNvNX/xYBp0ovncs8WyWZrOrpBNub/JfaMvbURyft8=


andrii-korotkov-verkada (approved):

andrii-korotkov-verkada:
Ah, looks like cloud doesn't support blobs https://docs.sonarsource.com/sonarqube-server/latest/project-administration/analysis-scope/#wildcard-examples. I can't find exact explanation what that means, but would try replacing `pkg/diff/testdata/**` with `pkg/diff/testdata/**/*` as well.

pjiang-dev:
> Ah, looks like cloud doesn't support blobs https://docs.sonarsource.com/sonarqube-server/latest/project-administration/analysis-scope/#wildcard-examples. I can't find exact explanation what that means, but would try replacing `pkg/diff/testdata/**` with `pkg/diff/testdata/**/*` as well.

Strange, tried a couple different wildcards and they don't exclude the testdata files

andrii-korotkov-verkada:
Did you try adding spaces around = ? Tho I swear it worked for me before just like it was in master.

andrii-korotkov-verkada:
Hm, now I can only think of two things:
* Sonar config usage is broken.
* Files need to be excluded individually.

zachaller:
I would imagine part of this issue is with Lists only due to not having merge keys, does RecursiveDifference only do that whole replacement on lists or does it do it for every object type? If it does it for every object type I am not sure that is the correct action but I will differ to @leoluz

pjiang-dev:
> I would imagine part of this issue is with Lists only due to not having merge keys, does RecursiveDifference only do that whole replacement on lists or does it do it for every object type? If it does it for every object type I am not sure that is the correct action but I will differ to @leoluz

This RecursiveDIfference only works on sets. This is coming from the kubernetes structured-diff library where the original DIfference comes from.

pjiang-dev:
There is a bug using this approach when manifest files have a field added by mutation webhooks:
Screenshot 2025-02-12 at 1 58 22 PM

Basically RecursiveDIfference will remove fields even if it is not an exact match. It works by removing subtrees only when there's an exact match at a non-leaf level.

Example:
Added by mutation webhooks set:
```
.metadata.creationTimestamp
.metadata.labels.peter-test
```

Managed Fields Set:
`.metadata.labels.app.kubernetes.io/instance`

In this case `.metadata.labels.peter-test` will be removed because it matches the field `.metadata.labels` inside the managed fields set, which is incorrect.

One option is to create a custom Difference function to remove fields at a more granular level.

andrii-korotkov-verkada:
https://github.com/kubernetes-sigs/structured-merge-diff/blob/master/fieldpath/set.go#L110
Maybe we can ask the community to implement the right variation of the function.

andrii-korotkov-verkada:
@leoluz, do you think we should implement a custom function or ask structured merge diff devs to do it? I feel like the 2nd is better, though would take some time.

andrii-korotkov-verkada:
Shall we only remove mfs for leaf nodes or something like that?

pkg/diff/diff.go
```diff
@@ -236,15 +236,15 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
 		}
 		if comparison.Added != nil && !comparison.Added.Empty() {
 			// exclude the added fields owned by this manager from the comparison
-			comparison.Added = comparison.Added.Difference(mfs)
+			comparison.Added = comparison.Added.RecursiveDifference(mfs)
```
andrii-korotkov-verkada, line 239: Or, I wonder if we can do something like
```
predictedLive.ExtractItems(plManagedFields, typed.WithAppendKeyFields())
```

andrii-korotkov-verkada: Or maybe we just remove every prefix of every managed field path.

pjiang-dev: I tried this as well, and it results to messed up outputs.

andrii-korotkov-verkada: Ah, nvm, I think we should use RecursiveDifference, but only with leaf mfs. What do you think?

andrii-korotkov-verkada: One more thought - also remove managed fields from the current live.

andrii-korotkov-verkada:
> In this case .metadata.labels.peter-test will be removed because it matches the field .metadata.labels inside the managed fields set, which is incorrect.

Why would `.metadata.labels` be a part of managed fields set if not all of it is managed?

andrii-korotkov-verkada:
One more thought - do a recursive difference, but only if the field is managed in both live and predicted live.

pkg/diff/diff.go
pjiang-dev: I think we can't use RecursiveDifference because it could incorrectly remove fields from mutation webhooks. 
However, yes, I think what we could do is keep Difference, but do a second pass to remove any leaf nodes that are still in managed fields that weren't caught by Difference.

andrii-korotkov-verkada:
```
// RecursiveDifference returns a Set containing elements which:
// * appear in s
// * do not appear in s2
//
// Compared to a regular difference,
// this removes every field **and its children** from s that is contained in s2.
//
// For example, with s containing `a.b.c` and s2 containing `a.b`,
// a RecursiveDifference will result in `a`, as the entire node `a.b` gets removed.
```
In the example
```
Added by mutation webhooks set:

.metadata.creationTimestamp
.metadata.labels.peter-test

Managed Fields Set:
.metadata.labels.app.kubernetes.io/instance

In this case .metadata.labels.peter-test will be removed because it matches the field .metadata.labels inside the managed fields set, which is incorrect.
```
I don't think `.metadata.labels.peter-test` would be removed, unless `.metadata.labels` is also in the managed field set (perhaps because predicted live only has `.metadata.labels.app.kubernetes.io/instance`.

pkg/diff/diff.go
pjiang-dev: I will try out this and also removing manged fields from current live and see if tit works

andrii-korotkov-verkada:
So IIUC in the example above predicted live should contain:
```
.metadata.creationTimestamp
.metadata.labels.peter-test
.metadata.labels.app.kubernetes.io/instance
```
But I don't see why `.metadata.labels` would be in the managed fields set.

andrii-korotkov-verkada:
Do I understand correctly, that the problem is with partially managed fields (with children) being considered as fully managed?

andrii-korotkov-verkada:
Can you also try just this modification, still with using Difference?

Now:
```
	if comparison.Added != nil && !comparison.Added.Empty() {
		typedPredictedLive = typedPredictedLive.RemoveItems(comparison.Added)
	}
```
Desired?
```
	if comparison.Added != nil && !comparison.Added.Empty() {
                // Might extract some key fields still used
                liveAddValues := typedLive.ExtractItems(comparison.Added, typed.WithAppendKeyFields()) 
		typedPredictedLive = typedPredictedLive.RemoveItems(comparison.Added)
                typedPredictedLive = typedPredictedLive.Merge(liveAddValues)
	}
```

pjiang-dev:
Here is what's happening exactly:

Before Difference() on Line 239
mfs:
```
.metadata.labels.app.kubernetes.io/instance
.spec.selector
.spec.template.metadata.labels.app.kubernetes.io/name
.spec.template.spec.containers[name="idle"]
.spec.template.spec.containers[name="idle"].image
.spec.template.spec.containers[name="idle"].name
.spec.template.spec.containers[name="idle"].env[name="FOO"]
.spec.template.spec.containers[name="idle"].env[name="FOO"].name
.spec.template.spec.containers[name="idle"].env[name="FOO"].value
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"]
.spec.template.spec.containers[name="idle"].ports[containerPort=9876,protocol="TCP"]
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].containerPort
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].name
.spec.template.spec.containers[name="idle"].ports[containerPort=9876,protocol="TCP"].containerPort
.spec.template.spec.containers[name="idle"].ports[containerPort=9876,protocol="TCP"].name
```

Comparison:
```
- Modified Fields:
.metadata.generation
.metadata.managedFields
- Added Fields:
.metadata.creationTimestamp
.metadata.labels.peter-test
.spec.template.spec.containers[name="idle"].env
.spec.template.spec.containers[name="idle"].env[name="FOO"]
.spec.template.spec.containers[name="idle"].env[name="FOO"].name
.spec.template.spec.containers[name="idle"].env[name="FOO"].value
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"]
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].containerPort
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].name
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].protocol
```

After Difference() on line 239
Comparison:
```
- Modified Fields:
.metadata.generation
.metadata.managedFields
- Added Fields:
.metadata.creationTimestamp
.metadata.labels.peter-test
.spec.template.spec.containers[name="idle"].env
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].protocol
```

The bug is that these two fields remain in comparison.Added even after the Difference function even though they are managed fields. If they are in comparison.Added they will be extracted from predictedLive and cause the diff to not show these fields. 
```
.spec.template.spec.containers[name="idle"].env
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].protocol
```

The cause is the Difference function will does a shallow difference and does not recursively check each child.

pjiang-dev:
> Can you also try just this modification, still with using Difference?
> 
> Now:
> 
> ```
> 	if comparison.Added != nil && !comparison.Added.Empty() {
> 		typedPredictedLive = typedPredictedLive.RemoveItems(comparison.Added)
> 	}
> ```
> 
> Desired?
> 
> ```
> 	if comparison.Added != nil && !comparison.Added.Empty() {
>                 // Might extract some key fields still used
>                 liveAddValues := typedLive.ExtractItems(comparison.Added, typed.WithAppendKeyFields()) 
> 		typedPredictedLive = typedPredictedLive.RemoveItems(comparison.Added)
>                 typedPredictedLive, _ = typedPredictedLive.Merge(liveAddValues)
> 	}
> ```

This one also did not make a difference. I think because we are still extracting those fields.

andrii-korotkov-verkada:
Great analysis!

> .spec.template.spec.containers[name="idle"].env
Looks like fields that are parents of other fields are not managed themselves.
We run into a tricky situation when they were added, but only some of their children are managed.
In such case, after removing managed fields, these kind of fields would still remain, so we shouldn't touch them.
In other words, we should remove a non-managed field if and only if all its children are managed.

> .spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].protocol
This is due to a key field not specified explicitly, but it can be derived from the path.
It's actually an easier case, since its parent is managed.
We should remove a non-managed field if its parent subpath is managed (e.g. `.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"]`).

andrii-korotkov-verkada:
The solution is:
We compute the leaves of added fields before the mfs subtraction loop. After the loop we go through the remaining added fields and add some fields to a new excluded field set:
* If it's a leaf, we add all of its subpaths to this field set.
* If any of its subpaths is managed, we add the path itself to this set.
Then we subtract this field set from the added fields.

pjiang-dev:
@andrii-korotkov-verkada I think this is a good solution. Basically it covers the two edge cases 
Case 1: Parent fields with managed children
.spec.template.spec.containers[name="idle"].env
(children like .env[name="FOO"] are managed)

Case 2: Derived fields with managed parents
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].protocol
(parent path is managed)

I can start implementing this and @leoluz can confirm if it makes sense

PR #688, opened 2025-02-18
fix: Server-side diff shows incorrect diffs for list related changes

pjiang-dev:
fixes https://github.com/argoproj/argo-cd/issues/21817

This PR is refactoring how we remove webhook mutation fields from server side diff. 
This is the following logic:
1. Construct a predicted live of ONLY argocd-controller managed fields from predicted live, this will also remove any webhook mutation fields as they do not have a field manager
2. Apply this predicted live from step 1 to live state. This will give us a new predicted live state without mutation webhook fields, as this will only show what was added/modified/removed from argocd-controller field manager.


pjiang-dev:
@leoluz @andrii-korotkov The last PR got closed so opening this new one. 
I tried out this strategy above. It seems to have fixed the issue the the env variables not being shown, but still the diff with 'ports' is still showing deleted:
Screenshot 2025-02-18 at 12 09 11 AM

Here is my findings:
1. After i apply (merge) the predictedLive to Live it ends up deleting the field. Removal of fields should never happen with this function. See https://pkg.go.dev/sigs.k8s.io/structured-merge-diff/typed#TypedValue.Merge
2. The reason I think that this happens starts from the issue that this field is not considered managed by argocd-controller:
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].protocol
3. Because that field is not managed by argocd-controller, we remove it from predicted live, causing a bad state.

Here is the fields managed by argocd-controller:
`.metadata.annotations
.metadata.labels
.metadata.annotations.kubectl.kubernetes.io/last-applied-configuration
.metadata.labels.app.kubernetes.io/instance
.spec.progressDeadlineSeconds
.spec.replicas
.spec.revisionHistoryLimit
.spec.selector
.spec.strategy.rollingUpdate
.spec.strategy.type
.spec.strategy.rollingUpdate.maxSurge
.spec.strategy.rollingUpdate.maxUnavailable
.spec.template.metadata.labels
.spec.template.metadata.labels.app.kubernetes.io/name
.spec.template.spec.dnsPolicy
.spec.template.spec.restartPolicy
.spec.template.spec.schedulerName
.spec.template.spec.securityContext
.spec.template.spec.terminationGracePeriodSeconds
.spec.template.spec.containers[name="idle"]
.spec.template.spec.containers[name="idle"].image
.spec.template.spec.containers[name="idle"].imagePullPolicy
.spec.template.spec.containers[name="idle"].name
.spec.template.spec.containers[name="idle"].ports
.spec.template.spec.containers[name="idle"].resources
.spec.template.spec.containers[name="idle"].terminationMessagePath
.spec.template.spec.containers[name="idle"].terminationMessagePolicy
.spec.template.spec.containers[name="idle"].env[name="FOO"]
.spec.template.spec.containers[name="idle"].env[name="FOO"].name
.spec.template.spec.containers[name="idle"].env[name="FOO"].value
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"]
.spec.template.spec.containers[name="idle"].ports[containerPort=9876,protocol="TCP"]
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].containerPort
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].name
.spec.template.spec.containers[name="idle"].ports[containerPort=9876,protocol="TCP"].containerPort
.spec.template.spec.containers[name="idle"].ports[containerPort=9876,protocol="TCP"].name
.spec.template.spec.containers[name="idle"].ports[containerPort=9876,protocol="TCP"].protocol`

And here is the predicted Live fields:
`.apiVersion
.kind
.metadata.creationTimestamp
.metadata.generation
.metadata.managedFields
.metadata.name
.metadata.namespace
.metadata.resourceVersion
.metadata.uid
.metadata.annotations.deployment.kubernetes.io/revision
.metadata.annotations.kubectl.kubernetes.io/last-applied-configuration
.metadata.labels.app.kubernetes.io/instance
.metadata.labels.peter-test
.spec.progressDeadlineSeconds
.spec.replicas
.spec.revisionHistoryLimit
.spec.selector
.spec.strategy.type
.spec.strategy.rollingUpdate.maxSurge
.spec.strategy.rollingUpdate.maxUnavailable
.spec.template.metadata.creationTimestamp
.spec.template.metadata.labels.app.kubernetes.io/name
.spec.template.spec.dnsPolicy
.spec.template.spec.restartPolicy
.spec.template.spec.schedulerName
.spec.template.spec.securityContext
.spec.template.spec.terminationGracePeriodSeconds
.spec.template.spec.containers[name="idle"]
.spec.template.spec.containers[name="idle"].image
.spec.template.spec.containers[name="idle"].imagePullPolicy
.spec.template.spec.containers[name="idle"].name
.spec.template.spec.containers[name="idle"].resources
.spec.template.spec.containers[name="idle"].terminationMessagePath
.spec.template.spec.containers[name="idle"].terminationMessagePolicy
.spec.template.spec.containers[name="idle"].env[name="FOO"]
.spec.template.spec.containers[name="idle"].env[name="FOO"].name
.spec.template.spec.containers[name="idle"].env[name="FOO"].value
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"]
.spec.template.spec.containers[name="idle"].ports[containerPort=9876,protocol="TCP"]
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].containerPort
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].name
.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].protocol
.spec.template.spec.containers[name="idle"].ports[containerPort=9876,protocol="TCP"].containerPort
.spec.template.spec.containers[name="idle"].ports[containerPort=9876,protocol="TCP"].name
.spec.template.spec.containers[name="idle"].ports[containerPort=9876,protocol="TCP"].protocol
.status.availableReplicas
.status.observedGeneration
.status.readyReplicas
.status.replicas
.status.updatedReplicas
.status.conditions[type="Available"]
.status.conditions[type="Progressing"]
.status.conditions[type="Available"].lastTransitionTime
.status.conditions[type="Available"].lastUpdateTime
.status.conditions[type="Available"].message
.status.conditions[type="Available"].reason
.status.conditions[type="Available"].status
.status.conditions[type="Available"].type
.status.conditions[type="Progressing"].lastTransitionTime
.status.conditions[type="Progressing"].lastUpdateTime
.status.conditions[type="Progressing"].message
.status.conditions[type="Progressing"].reason
.status.conditions[type="Progressing"].status
.status.conditions[type="Progressing"].type`

What is very strange is this field `.spec.template.spec.containers[name="idle"].ports[containerPort=9876,protocol="TCP"].protocol `is considered managed by argocd-controller, However this protocol field is not:
`.spec.template.spec.containers[name="idle"].ports[containerPort=3000,protocol="TCP"].protocol`

andrii-korotkov-verkada:
> Does anyone have any insight on these protocol fields and why some are managed by argocd and others not?

I think it depends on whether it's specified explicitly or Argo CD relies on default value for the key field.
Looks like it can manage the parent paths explicitly, but not all the children if they were not specified explicitly.

My solution covers this case:
> The solution is:
> We compute the leaves of added fields before the mfs subtraction loop. After the loop we go through the remaining  added fields and add some fields to a new excluded field set:
> * If it's a leaf, we add all of its subpaths to this field set.
> * If any of its subpaths is managed, we add the path itself to this set.

pjiang-dev:
@andrii-korotkov-verkada Thanks. The root issue is that when we don't specify this 'protocol' field when adding a new container port, kubernetes does not consider this a managed field and this function will try to remove it when doing ssd which is causing the issues.

I just added some code to specifically look for this protocol field and try to remove it when removing non-argocd-controller fields and it fixes the issue. But hard coding this is probably not the right way. 

The other way is to keep the current removeWebhookMutation and use andrii's solution for computing managed fields before we remove.

pkg/diff/diff.go
```diff
@@ -220,57 +222,59 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
...
-			comparison.Added = comparison.Added.Difference(mfs)
-		}
-		if comparison.Modified != nil && !comparison.Modified.Empty() {
-			// exclude the modified fields owned by this manager from the comparison
-			comparison.Modified = comparison.Modified.Difference(mfs)
-		}
-		if comparison.Removed != nil && !comparison.Removed.Empty() {
-			// exclude the removed fields owned by this manager from the comparison
-			comparison.Removed = comparison.Removed.Difference(mfs)
+		if mfEntry.Manager == fieldManagerName {
```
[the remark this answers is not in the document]
pjiang-dev: manager is indeed argocd-controller which is what we want to check for. I am using it in my change now

pjiang-dev:
@leoluz @andrii-korotkov-verkada 
I've found some more details about the root cause and i updated with a potential fix:

The Merge function was not the issue. The issue is that after removing the protocol field, the predictedLive schema breaks:
`.spec.template.spec.containers[name="idle"].ports: element 1: associative list with keys has an element that omits key field "protocol" (and doesn't have default value)`
There are certain fields like protocol that must be in the schema and if they are deleted it leaves "dangling" partial structures that are not valid by themselves

The solution i have written and tested is to validate the removal of these non argocd-controller fields, if they don't break the schema on removal, then we can safely remove it and if it breaks the schema then we will leave it. This should also cause any other default fields like protocol to be preserved in this function.

Testing it so far and everything is working as expected. Let me know your thoughts

pkg/diff/diff.go
```diff
@@ -37,6 +38,7 @@ const (
 	couldNotMarshalErrMsg       = "Could not unmarshal to object of type %s: %v"
 	AnnotationLastAppliedConfig = "kubectl.kubernetes.io/last-applied-configuration"
 	replacement                 = "++++++++"
+	fieldManagerName            = "argocd-controller"
```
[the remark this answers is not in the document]
pjiang-dev: I've removed any mention of the argocd controller in gitops engine now. Also i am using the same manager that we use in the `serverSideDryRunner ` so the manager name will always be consistent.

pjiang-dev:
Trying to exclude the testdata manifests but nothing seems to work @andrii-korotkov-verkada

leoluz (changes requested):
Please add more unit tests based on our conversation.

pjiang-dev:
Added another unit test to include adding another container from kubectl manager using `kubectl apply` and passes this test.

The test validates that adding another container using `kubectl apply` does not break the server side diff.

The current implementation handles this case in the following manner:
1. When removing non argocd-controller fields, we find all non argocd fields, including those added by kubectl 
2. Inside the new function `safelyRemoveFieldsSet` removing this field owned by kubectl is still a valid schema: 
` .spec.template.spec.containers[name="new-container"].ports[containerPort=8080,protocol="TCP"]`
3. The field is removed and then merged with live to produce the new predcitedLive which will only show fields belonging to argocd-controller.

pkg/diff/diff.go
leoluz, line 1: Please remove all Argo CD mentions from this file (variable names and comments) as  gitops-engine is a library.

pkg/diff/diff.go
```diff
@@ -199,7 +199,7 @@ func serverSideDiff(config, live *unstructured.Unstructured, opts ...Option) (*D
 // managedFields. All fields under this condition will be reverted with their state
 // from live. If the given predictedLive does not have the managedFields, an error
 // will be returned.
```
leoluz, line 201: Please update the doc, describing the new logic being applied and how the given `manager` parameter will affect the results.

leoluz (changes requested):

pkg/diff/diff.go
```diff
@@ -197,9 +197,11 @@ func serverSideDiff(config, live *unstructured.Unstructured, opts ...Option) (*D
...
 // changes in predictedLive fields not associated with any manager in the
 // managedFields. All fields under this condition will be reverted with their state
-// from live. If the given predictedLive does not have the managedFields, an error
-// will be returned.
-func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkParser *managedfields.GvkParser, _ string) (*unstructured.Unstructured, error) {
+// from live. The resulting diff after this operation will only contain fields that were managed
+// by the provided 'manager' parameter. Any fields not managed by the specified
+// manager will also will be reverted with their state from live.
+// If the given predictedLive does not have the managedFields, an error will be returned.
+func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkParser *managedfields.GvkParser, manager string) (*unstructured.Unstructured, error) {
```
leoluz, line 204: > Webhook mutations are identified by finding changes in predictedLive fields not associated with any manager in the managedFields.

This isn't the logic that we are executing anymore right?

leoluz (changes requested):

leoluz (changes requested):

pkg/diff/diff.go
```diff
@@ -281,6 +268,28 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
 	return &unstructured.Unstructured{Object: pl}, nil
 }
 
+func safelyRemoveFieldsSet(predictedLive *typed.TypedValue, fieldsToRemove *fieldpath.Set) *fieldpath.Set {
```
leoluz, line 271: Please add the godoc describing what "safe" means in this context.

leoluz (changes requested):

pkg/diff/diff.go
```diff
@@ -197,9 +197,11 @@ func serverSideDiff(config, live *unstructured.Unstructured, opts ...Option) (*D
 // changes done by mutation webhooks. Webhook mutations are identified by finding
 // changes in predictedLive fields not associated with any manager in the
 // managedFields. All fields under this condition will be reverted with their state
-// from live. If the given predictedLive does not have the managedFields, an error
-// will be returned.
-func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkParser *managedfields.GvkParser, _ string) (*unstructured.Unstructured, error) {
+// from live. The resulting diff after this operation will only contain fields that were managed
```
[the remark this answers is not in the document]
pjiang-dev: Thats right. I'll update this doc

pkg/diff/diff.go
pjiang-dev: That statement is still true in the new logic, but we are now removing the webhook mutations in a different way

pjiang-dev: Updated doc again to clarify

leoluz (approved):
LGTM

ahmetb:
Any ideas when this fix would be picked up by https://github.com/argoproj/gitops-engine/commits/release-2.14 branch? I see newer commits landing there but this one seems to be skipped.

pjiang-dev:
Thanks for pointing this out @ahmetb i'm going to cherry pick this to 2.14

jeremych1000:
@pjiang-dev do you know if this fixes https://github.com/argoproj/argo-cd/issues/18839 as well? :)

PR #699, opened 2025-03-12
fix: Fix checking dryRun when using Server Side Apply

pjiang-dev:
Fixes [ #21858](https://github.com/argoproj/argo-cd/issues/21858)

We did not properly check the correct dryRun flag before using ServerSideApply. We had previously checked only the sync context instead of passing the actual dryRun flag

During ArgoCD sync we always apply all resources in dry-run mode to make sure that all yamls are valid. In this case the
`sc.dryRun=False, dryRun=True`
Once we hit this case, we are essentially doing `kubectl apply --serverSide=true --dryRun=client` which leads to unintended results and actually wet applies resources: https://github.com/kubernetes/kubectl/blame/deef152435a8265c9cc91ff12e43c87132dbb0a8/pkg/cmd/apply/apply.go#L573

Following the bug reproduce steps:
Before fix:
Create application with resources across multiple sync waves (or hooks). I use a job in wave -1 in this case).
Sync it with server-side apply.
The sync will fail (as the job fails): however, the Deployment from a later wave was still synced.

After the fix:
Create application with resources across multiple sync waves (or hooks). I use a job in wave -1 in this case).
After performing sync with serverSideApply:
The sync fails(as the job fails). The job properly degrades, and the subsequent deployment (apply-to-sync-later) is still out of sync and doesn't apply.
Screenshot 2025-03-12 at 2 29 24 PM

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -1013,12 +1013,13 @@ func (sc *syncContext) ensureCRDReady(name string) error {
 	})
 }
 
-func (sc *syncContext) shouldUseServerSideApply(targetObj *unstructured.Unstructured) bool {
+func (sc *syncContext) shouldUseServerSideApply(targetObj *unstructured.Unstructured, dryRun bool) bool {
 	// if it is a dry run, disable server side apply, as the goal is to validate only the
 	// yaml correctness of the rendered manifests.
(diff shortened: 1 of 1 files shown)


leoluz (approved):
LGTM

PR #703, opened 2025-03-13
chore: add unit test for ssa with dryRun

pjiang-dev:
Adds unit test for https://github.com/argoproj/gitops-engine/commit/762f9b70f3056381aa782a4ea310a57919380d49
Covers the cases where if we are doing a dry run, we should never use server side apply.


leoluz (approved):
LGTM

PR #708, opened 2025-04-08
Add option to skip the dryrun from the sync context

Nheijmink19:
This is a start to solve the issue that the validation fails on creating resources in a not yet existing namespace.
https://github.com/argoproj/argo-cd/issues/21788

By setting the skipDryRun you can skip the validation in your sync action allowing the resources that can already be created (the namespace) to be created and the other resources in another sync wave. 

Also added some typing to some params.

If oke I will join the meeting next thursday to explain the requested change.

### DIFF
--- a/pkg/sync/common/types.go
+++ b/pkg/sync/common/types.go
@@ -20,6 +20,8 @@ const (
 
 	// Sync option that disables dry run in resource is missing in the cluster
 	SyncOptionSkipDryRunOnMissingResource = "SkipDryRunOnMissingResource=true"
+	// Sync option that disables dry run for applying resources
+	SyncOptionSkipDryRun = "SkipDryRun=false"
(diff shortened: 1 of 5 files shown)


pkg/sync/common/types.go
```diff
@@ -20,6 +20,8 @@ const (
 
 	// Sync option that disables dry run in resource is missing in the cluster
 	SyncOptionSkipDryRunOnMissingResource = "SkipDryRunOnMissingResource=true"
+	// Sync option that disables dry run for applying resources
+	SyncOptionSkipDryRun = "SkipDryRun=true"
```
alexmt, line 24: 👍

alexmt (approved):
Thanks for updating the sync option value and implementing a test. Looking forward for Argo CD PR!

LGTM

PR #710, opened 2025-04-12
fix: go list is not working due to k8s.io/externaljwt dependency

pasha-codefresh:
Getting error when executing `go list -modfile=~/go/src/github.com/argoproj/gitops-engine/go.mod -m -json -mod=mod all` locally it fails with 
Screenshot 2025-04-12 at 18 36 38

Goland executing it in order to load dependencies list

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -119,6 +119,7 @@ replace (
 	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.32.2
 	k8s.io/dynamic-resource-allocation => k8s.io/dynamic-resource-allocation v0.32.2
 	k8s.io/endpointslice => k8s.io/endpointslice v0.32.2
+	k8s.io/externaljwt => k8s.io/externaljwt v0.32.2
 	k8s.io/kms => k8s.io/kms v0.32.2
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -119,6 +119,7 @@ replace (
 	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.32.2
 	k8s.io/dynamic-resource-allocation => k8s.io/dynamic-resource-allocation v0.32.2
 	k8s.io/endpointslice => k8s.io/endpointslice v0.32.2
+	k8s.io/externaljwt => k8s.io/externaljwt v0.32.2
```
leoluz, line 122: It seems that something else is off here. Why we have this library only listed in the `replace` section and not in the `require` section?

leoluz (changes requested):

leoluz (approved):
As discussed with @pasha-codefresh , the problem is that `k8s.io/externaljwt` is a dependency introduced by `k8s.io/kubernetes` that shouldn't be used as a library:
https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-505725449

PR #712, opened 2025-04-13
feat: Enable SkipDryRunOnMissingResource sync option on Application level

pasha-codefresh:
Related to https://github.com/argoproj/argo-cd/issues/21790

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -194,6 +194,12 @@ func WithSkipDryRun(skipDryRun bool) SyncOpt {
 	}
 }
 
+func WithSkipDryRunOnMissingResource(skipDryRunOnMissingResource bool) SyncOpt {
+	return func(ctx *syncContext) {
+		ctx.skipDryRunOnMissingResource = skipDryRunOnMissingResource
+	}
+}
+
 func WithServerSideApply(serverSideApply bool) SyncOpt {
 	return func(ctx *syncContext) {
 		ctx.serverSideApply = serverSideApply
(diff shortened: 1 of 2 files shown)


Nheijmink19:
couldn't you just use the new skipDryrun functionality? that one is on application scope? (not yet available in argoCD, but implemented in the git-ops engine)

pkg/sync/sync_context.go
```diff
@@ -818,11 +825,18 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
...
+
 		if err != nil {
 			switch {
-			case apierrors.IsNotFound(err) &&
-				((task.targetObj != nil && resourceutil.HasAnnotationOption(task.targetObj, common.AnnotationSyncOptions, common.SyncOptionSkipDryRunOnMissingResource)) ||
-					sc.hasCRDOfGroupKind(task.group(), task.kind())):
+			case apierrors.IsNotFound(err) && shouldSkipDryRunOnMissingResource():
 				// Special case for custom resources: if CRD is not yet known by the K8s API server,
 				// and the CRD is part of this sync or the resource is annotated with SkipDryRunOnMissingResource=true,
 				// then skip verification during `kubectl apply --dry-run` since we expect the CRD
```
reggie-k, line 842: Does this comment now belong to the newly created `shouldSkipDryRunOnMissingResource` func above?

pasha-codefresh: Not really, it can stay inside case condition

reggie-k (approved):

PR #716, opened 2025-05-04
chore: Use watch APIs to list k8s resources

andrii-korotkov:
Use watch APIs to list resources instead of paginated regular APIs.

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -30,6 +31,7 @@ import (
 	"k8s.io/client-go/tools/pager"
 	watchutil "k8s.io/client-go/tools/watch"
 	"k8s.io/client-go/util/retry"
+	"k8s.io/client-go/util/watchlist"
 	"k8s.io/klog/v2/textlogger"
 	"k8s.io/kubectl/pkg/util/openapi"
 
@@ -248,7 +250,9 @@ type clusterCache struct {
 	openAPISchema               openapi.Resources
 	gvkParser                   *managedfields.GvkParser
 
-	respectRBAC int
+	respectRBAC                  int
+	listResourcesUsingWatchAPI   atomic.Int32
+	listResourcesUsingRegularAPI atomic.Int32
 }
 
 type clusterCacheSync struct {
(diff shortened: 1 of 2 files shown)


blakepettersson:
Is there some way this can be benchmarket? Like what are the expected performance improvements with this method (I imagine they are large, but ballpark numbers)?

andrii-korotkov:
> Is there some way this can be benchmarket? Like what are the expected performance improvements with this method (I imagine they are large, but ballpark numbers)?

I'll try to think of something.

pkg/cache/cluster.go
```diff
@@ -600,10 +604,26 @@ func (c *clusterCache) listResources(ctx context.Context, resClient dynamic.Reso
 			listRetry = retry.DefaultRetry
 		}
 
+		if opts.ResourceVersion == "" {
+			opts.ResourceVersion = "0"
+		}
```
crenshaw-dev, line 609: Why is this necessary?

pkg/cache/cluster.go
```diff
@@ -248,7 +250,9 @@ type clusterCache struct {
 	openAPISchema               openapi.Resources
 	gvkParser                   *managedfields.GvkParser
 
-	respectRBAC int
+	respectRBAC                  int
+	listResourcesUsingWatchAPI   atomic.Int32
+	listResourcesUsingRegularAPI atomic.Int32
```
crenshaw-dev, line 255: I think maybe we shouldn't do this, because listResources gets called a lot over the lifetime of a controller run (i.e. every time the cache expires or there's a retry due to an error). I don't think it's worth it just for the test. And maybe we can find a different way to test.

pkg/cache/cluster_test.go
```diff
@@ -189,41 +190,80 @@ func Benchmark_sync(t *testing.B) {
...
-	require.NoError(t, err)
+	originalFeatureGates := clientfeatures.FeatureGates()
 
-	cluster.lock.Lock()
-	defer cluster.lock.Unlock()
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.listUsingWatchAPIs {
+				// Enable WatchListClient in particular. Setting via env variable here is too late.
+				clientfeatures.ReplaceFeatureGates(AlwaysEnabledGates{})
```
crenshaw-dev, line 220: Looks like this enables every feature gate that can be enabled. Can we instead enable just the watch-list gate?

crenshaw-dev (changes requested):

pkg/cache/cluster.go
andrii-korotkov: Required by the list options construction function's code, otherwise it would return false as success.

pkg/cache/cluster_test.go
andrii-korotkov: It's a bit more complicated, and it's just for this test, so I'd rather keep this.

pkg/cache/cluster.go
andrii-korotkov: Agree that it's not ideal, I'll think how to test better.

pkg/cache/cluster_test.go
andrii-korotkov: Although shouldn't be too hard to construct gates more precisely.

pkg/cache/cluster.go
rumstead: https://github.com/argoproj/gitops-engine/pull/617#discussion_r1723570601

Referenced from https://ahmet.im/blog/kubernetes-list-performance/?utm_id=FAUN_Kaptain485_Link_title#list-api-performance-characteristics

PR #718, opened 2025-05-09
Fix generating incorrect names for Jobs when `generateName` is used

pnovitskiy:
Discovered a case when processing of `generateName` produces incorrect results.

For example if you have revision `1.1515.1` it's being cut a `1.1515.` and combined with suffix that starts with `-` character produces name that does not comply with RFC1123.

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/sync/sync_context.go
```diff
@@ -754,7 +754,7 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
 				if targetObj.GetName() == "" {
 					var syncRevision string
 					if len(sc.revision) >= 8 {
-						syncRevision = sc.revision[0:7]
+						syncRevision = strings.Trim(sc.revision[0:7], ".")
```
GMartinez-Sisti, line 757: Semver also allows dashes (`-`) and plus (`+`), I would trim for that too.

pnovitskiy: Dashes should be fine. Let's keep only letters, numbers and `-`.

PR #727, opened 2025-06-02
feat: auto migrate kubectl-client-side-apply fields for SSA

pjiang-dev:
fixes https://github.com/argoproj/argo-cd/issues/23214

When performing a Sync with ServerSideApply we will check if there is a `kubectl-client-side-apply` manager and run a client side apply with the field manager `kubectl-client-side-apply`. This will make the` last-applied-confiugration` field to be owned by `kubectl-client-side-apply` and then once the Sync with ServerSideApply executes it will trigger the auto-migration of fields to argocd-controller which is described in https://github.com/kubernetes/kubernetes/pull/112905

We need this in cases where a user created their resource with `kubectl apply -f` in the command line then later used argocd to manage it with SSA.

There will be a PR in argo-cd to support configurable field managers to migration from as well as option to disable this feature entirely

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -1096,11 +1096,11 @@ func (sc *syncContext) needsClientSideApplyMigration(liveObj *unstructured.Unstr
 	return false
 }
 
-// performClientSideApplyMigration does a client-side-apply with the specified field manager.
+// performClientSideApplyMigration performs a client-side-apply using the specified field manager.
 // This moves the 'last-applied-configuration' field to be managed by the specified manager.
-// Then the next time server-side apply happens, kubernetes automatically migrates all fields from the manager
-// that owns 'last-applied-configuration' to the manager that uses server-side apply.
-// kubectl-client-side-apply is used as the default manager if one is not specified.
+// The next time server-side apply is performed, kubernetes automatically migrates all fields from the manager
+// that owns 'last-applied-configuration' to the manager that uses server-side apply. This will remove the
+// specified manager from the resources managed fields. 'kubectl-client-side-apply' is used as the default manager.
 func (sc *syncContext) performClientSideApplyMigration(targetObj *unstructured.Unstructured, fieldManager string) error {
 	sc.log.WithValues("resource", kubeutil.GetResourceKey(targetObj)).V(1).Info("Performing client-side apply migration step")
 


pkg/sync/sync_context.go
```diff
@@ -1075,6 +1118,15 @@ func (sc *syncContext) applyObject(t *syncTask, dryRun, validate bool) (common.R
 	shouldReplace := sc.replace || resourceutil.HasAnnotationOption(t.targetObj, common.AnnotationSyncOptions, common.SyncOptionReplace)
 	force := sc.force || resourceutil.HasAnnotationOption(t.targetObj, common.AnnotationSyncOptions, common.SyncOptionForce)
 	serverSideApply := sc.shouldUseServerSideApply(t.targetObj, dryRun)
+
+	// Check if we need to perform client-side apply migration for server-side apply
+	if serverSideApply && !dryRun && sc.needsClientSideApplyMigration(t.liveObj) {
+		err = sc.performClientSideApplyMigration(t.targetObj)
+		if err != nil {
+			sc.log.WithValues("resource", kubeutil.GetResourceKey(t.targetObj)).Error(err, "Failed to perform client-side apply migration")
```
crenshaw-dev, line 1126: Is this failure mode loud enough? Should we instead fail the operation?

pjiang-dev: I decided here not to fail the operation as i didn't want to prevent the entire Sync from happening in case the client-side apply fails. What do you think?

pkg/sync/sync_context.go
```diff
@@ -1059,6 +1059,49 @@ func (sc *syncContext) shouldUseServerSideApply(targetObj *unstructured.Unstruct
...
+
+	managedFields := liveObj.GetManagedFields()
+	if len(managedFields) == 0 {
+		return false
+	}
+
+	for _, field := range managedFields {
+		if field.Manager == "kubectl-client-side-apply" && field.Operation == metav1.ManagedFieldsOperationUpdate {
+			return true
+		}
```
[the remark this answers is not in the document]
pjiang-dev: Yes, this would be the most common case and also this is the only manager where we already know the name of the manager beforehand. For other controllers users will have to manually handle those cases. Maybe we can document that somewhere in another PR

blakepettersson:
@pjiang-dev @crenshaw-dev could we not do something similar as kubernetes/kubernetes#124191 where we'd force ownership of all fields, given that we opt-in to this behavior by adding another sync option? Something like `OverwriteFieldManagers=true`?

pjiang-dev:
> @pjiang-dev @crenshaw-dev could we not do something similar as [kubernetes/kubernetes#124191](https://github.com/kubernetes/kubernetes/pull/124191) where we'd force ownership of all fields, given that we opt-in to this behavior by adding another sync option? Something like `OverwriteFieldManagers=true`?

@blakepettersson That PR works by manipulating managed fields and adds the '`last-applied-configuration`' to all field managers before doing SSA. Kubernetes recommends against doing so and also i don't know what side affects this may have. I think its safer to use what k8s already provides. Also some users may want to keep certain field managers.

However, i do think what we can do is provide an additional annotation with SSA that allows users to specify their own field manager to migrate fields. This would replace the 'kubectl-client-side-apply' field manager in this PR with their own with something like this:
`argocd.argoproj.io/sync-options: ServerSideApply=true, OverwriteFieldManager=<your-field-manager>`
this way they can have the flexibility to migrate whichever field manager they choose. We would still want to keep `kubectl-client-side-apply` as the default for this though.

blakepettersson:
>However, i do think what we can do is provide an additional annotation with SSA that allows users to specify their own field manager to migrate fields. This would replace the 'kubectl-client-side-apply' field manager in this PR with their own with something like this

Sounds good to me! Could we have that option for multiple field managers, e.g 

```
 argocd.argoproj.io/sync-options: ServerSideApply=true, OverwriteFieldManagers=<your-field-manager-1>,<your-field-manager-1>
```

(the default still being `kubectl-client-side-apply`)

pjiang-dev:
> provide an additional annotation with SSA that allows users to specify their own field manager to migrate fields. This would replace the 'kubectl-client-side-apply' field manager in this PR with their own with something like this:
> `argocd.argoproj.io/sync-options: ServerSideApply=true, OverwriteFieldManager=<your-field-manager>`
> this way they can have the flexibility to migrate whichever field manager they choose. We would still want to keep `kubectl-client-side-apply` as the default for this though.

Unfortunately, we cannot do multiple field managers at once because 'last-applied-configuration' can only belong to one field manager at a time without manipulating managed fields.

pkg/sync/sync_context.go
leoluz: I think we should return this error to the user instead of "hiding" it in the logs. Make sure that the error propagates in the sync context.

leoluz (changes requested):

pjiang-dev:
For this PR i will leave as is and create a follow up PR to make the field manager to migrate fields configurable to the user

leoluz (changes requested):
> For this PR i will leave as is and create a follow up PR to make the field manager to migrate fields configurable to the user

I think the only thing missing in this PR is the ability to turn this feature on/off

leoluz (approved):
LGTM

PR #728, opened 2025-06-04
Feature skip dryrun in all cases

Nheijmink19:
This is an extention to solve the issue that the validation fails on creating resources in a not yet existing namespace.
https://github.com/argoproj/argo-cd/issues/21788
and a followup to:
https://github.com/argoproj/gitops-engine/pull/708

By setting the skipDryRun you can skip the validation in your sync action allowing the resources that can already be created (the namespace) to be created and the other resources in another sync wave.
In the old situation it only worked for CRD's which made it redundant to skipDryRunOnMissingResource. By moving it out of the err!=nil check it can be used by all type of resources.

If oke I will join the meeting next thursday to explain the requested change.

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -834,14 +834,15 @@ func (sc *syncContext) getSyncTasks() (_ syncTasks, successful bool) {
 				sc.hasCRDOfGroupKind(task.group(), task.kind())
 		}
 
-		if sc.skipDryRun && task.liveObj == nil {
+		switch {
+		case sc.skipDryRun && task.liveObj == nil:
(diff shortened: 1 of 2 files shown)


agaudreault:
Closing due to https://github.com/argoproj/gitops-engine/pull/730

Will be replace by creating the namespace before the DryRun if the namespace does not exist. 

Eventually, a new option should be added to perform the dry run on the current wave instead of on all the resource before the sync.

Nheijmink19:
@agaudreault I did create something like that in the past:
https://github.com/argoproj/gitops-engine/pull/664/files
I am not sure if my implementation is the way you planned on building it, but if it is you already have a start.

PR #730, opened 2025-06-05
Revert "Add option to skip the dryrun from the sync context (#708)"

crenshaw-dev:
I changed my mind and now think this is more of a hatchet solution than the scalpel we need to solve https://github.com/argoproj/argo-cd/issues/21788.

Disabling dry runs entirely means you completely lose validation indefinitely. That's an aggressive solution to a problem that likely only presents itself once in the lifecycle of an app.

Talked with @agaudreault, and we think we can solve the problem with a lighter touch.

This change hasn't been released in any Argo CD version yet, so it's safe to revert.

@alexmt since you merged the PR initially, lmk if you'd like to discuss further!

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -13,6 +13,7 @@ linters:
     - unparam
     - usestdlibvars
     - whitespace
+    - wrapcheck
   settings:
     gocritic:
       disabled-checks:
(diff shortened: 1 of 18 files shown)


agaudreault (approved):

alexmt:
@crenshaw-dev If I was a user I don't think I would ever choose to have this sync option on resource level. However I totally can see how someone might want to sync app once with this option (via starting sync operation with sync option ). So I still think it was useful.

PR #734, opened 2025-06-12
Require pruning confirmation syncoption

MrFreezeex:


pjiang-dev (approved):

MrFreezeex:
(note that I only changed the second commit title that was point at the wrong sync option before :x)

agaudreault (changes requested):
code LGTM. Missing unit test and waiting for answers in the linked argo CD PR

PR #738, opened 2025-06-17
feat: allow list items to be processed in parallel

shady-canva:
This PR introduces a new setting which allows items returned by `ListPager` to be processed in parallel.

Currently, items are processed sequentially per resource. This is a problem if `populateResourceInfoHandler` is expensive. We internally have a custom resource which takes 1.5 milliseconds in the `populateResourceInfoHandler` function due to a custom lua healthcheck script and a few ignored labels which are removed during diff normalisation. With over 50,000 of these custom resources, that's a 75 second overhead.

With a parallelism of 8, we've seen a page of 500 items go from ~750ms to ~100ms. We've also seen the processing of a full list complete at roughly the same time as `ListPager` completes. Previously, processing of pages constantly lagged behind the retrieval.

`gobench` results:

```
BenchmarkSync/weight=1,overhead=100μs-12          	      68	  16072692 ns/op
BenchmarkSync/weight=2,overhead=100μs-12          	     139	   8585231 ns/op
BenchmarkSync/weight=4,overhead=100μs-12          	     242	   4905336 ns/op
BenchmarkSync/weight=8,overhead=100μs-12          	     302	   3939709 ns/op

BenchmarkSync/weight=1,overhead=500μs-12          	      19	  63097167 ns/op
BenchmarkSync/weight=2,overhead=500μs-12          	      38	  31135898 ns/op
BenchmarkSync/weight=4,overhead=500μs-12          	      75	  15546984 ns/op
BenchmarkSync/weight=8,overhead=500μs-12          	     141	   8420676 ns/op
```

Before and after:

image


pkg/cache/cluster.go
```diff
@@ -164,15 +166,16 @@ type ListRetryFunc func(err error) bool
...
-		resources:          make(map[kube.ResourceKey]*Resource),
-		nsIndex:            make(map[string]map[kube.ResourceKey]*Resource),
-		config:             config,
+		settings:                Settings{ResourceHealthOverride: &noopSettings{}, ResourcesFilter: &noopSettings{}},
+		apisMeta:                make(map[schema.GroupKind]*apiMeta),
+		eventMetaCh:             nil,
+		listPageSize:            defaultListPageSize,
+		listPageBufferSize:      defaultListPageBufferSize,
+		listSemaphore:           semaphore.NewWeighted(defaultListSemaphoreWeight),
+		listItemSemaphoreWeight: defaultListItemSemaphoreWeight,
```
FourFifthsCode, line 175: this is great that it maintains existing behavior by default

FourFifthsCode (approved):
very elegant solution!

PR #744, opened 2025-07-05
feat(syncwaves): use binary tree ordering for sync waves

SebastienFelix:


alexec:
Is this trying to address the same issue as https://github.com/argoproj/gitops-engine/pull/514 ?

SebastienFelix:
@alexec : kind of, I guess!
The main goal of this feature is to avoid wasting time by waiting for non-related resources to be synced before proceeding a given resource.

I thought about defining the dependencies relation graph the way you did it but I thought that managing the circular dependencies would be too complicated.

The way I did it in this PR completely avoid this issue and avoids adding more complexity with another logical layer.

WDYT?

crenshaw-dev:
I think folks' main complaint about sync waves is the user experience, not the lack of support for sequential steps within parallel waves. (There's also the request for ordering within a broader context than a single app, but that's a much harder nut to crack).

crenshaw-dev:
I think I'd find built-in DAG support with dependsOn within a single app context more attractive than a binary tree. It would solve parallelism and UX struggles.

But I think before bothering with either, I'd really want to interrogate your specific use case in detail. I think most use cases (aside from mutating webhooks) can be solved with retries and eventual consistency.

SebastienFelix:
Sure, we can discuss that outside here as it could probably be smoother with a quick call!

PR #745, opened 2025-07-10
fix: invalidate cluster cache after sync operations

pjiang-dev:
fixes https://github.com/argoproj/argo-cd/issues/20458

There have been several times after a sync that the diff shows the live resource before the sync, causing incorrect diffs. This change will ensure if there is a sync with a valid task (non no-op) then we invalidate the cache. We only invalidate cache for resources that were modified to save on operational cost
This is guaranteeing fresh data for subsequent diff calculations

This could also help/fix an issue with CRDs as previously after a CRD schema changes and is synced, we run into a diff error because diff is using stale schema

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -84,6 +84,13 @@ func (e *gitOpsEngine) Sync(ctx context.Context,
 		return nil, fmt.Errorf("failed to diff objects: %w", err)
 	}
 	opts = append(opts, sync.WithSkipHooks(!diffRes.Modified))
+
+	// Add cache invalidation callback to invalidate cache for modified resources after sync
+	opts = append(opts, sync.WithCacheInvalidationCallback(func() {
+		// Invalidate the entire cache to ensure consistency
+		e.cache.Invalidate()
+	}))
+
 	syncCtx, cleanup, err := sync.NewSyncContext(revision, result, e.config, e.config, e.kubectl, namespace, e.cache.GetOpenAPISchema(), opts...)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create sync context: %w", err)
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -74,6 +74,13 @@ func (e *gitOpsEngine) Sync(ctx context.Context,
 	namespace string,
 	opts ...sync.SyncOpt,
 ) ([]common.ResourceSyncResult, error) {
+	// Ensure cache is synced before getting managed live objects
+	// This forces a refresh if the cache was invalidated
+	err := e.cache.EnsureSynced()
+	if err != nil {
+		return nil, fmt.Errorf("failed to ensure cache is synced: %w", err)
+	}
+
 	managedResources, err := e.cache.GetManagedLiveObjs(resources, isManaged)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get managed live objects: %w", err)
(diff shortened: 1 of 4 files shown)


pkg/engine/engine.go
```diff
@@ -74,6 +74,13 @@ func (e *gitOpsEngine) Sync(ctx context.Context,
 	namespace string,
 	opts ...sync.SyncOpt,
 ) ([]common.ResourceSyncResult, error) {
+	// Ensure cache is synced before getting managed live objects
+	// This forces a refresh if the cache was invalidated
+	err := e.cache.EnsureSynced()
+	if err != nil {
+		return nil, fmt.Errorf("failed to ensure cache is synced: %w", err)
```
leoluz, line 81: maybe this to facilitate the understanding while reading the logs?
"error during sync: failed to ensure cache is synced"

leoluz (changes requested):
Please check my comments.

pkg/engine/engine.go
```diff
@@ -84,6 +91,15 @@ func (e *gitOpsEngine) Sync(ctx context.Context,
...
 	}
 	opts = append(opts, sync.WithSkipHooks(!diffRes.Modified))
+
+	// Add cache invalidation callback to invalidate cache for modified resources after sync
+	opts = append(opts, sync.WithCacheInvalidationCallback(func(modifiedResources []kube.ResourceKey) {
+		// Only invalidate the specific resources that were modified
+		if len(modifiedResources) > 0 {
+			e.cache.InvalidateResources(modifiedResources)
+		}
+	}))
```
leoluz, line 101: Why would we need to invalidate the cache on every sync if the cluster cache is based on resource watches? I feel that the root problem is elsewhere. WDYT?

pkg/sync/sync_context.go
```diff
@@ -557,6 +569,15 @@ func (sc *syncContext) Sync() {
...
 		sc.setOperationPhase(common.OperationSucceeded, "successfully synced (no more tasks)")
+
+		// Invalidate cache after successful sync
+		if sc.cacheInvalidationCallback != nil {
+			modifiedResources := make([]kubeutil.ResourceKey, 0, len(sc.syncRes))
+			for _, result := range sc.syncRes {
+				modifiedResources = append(modifiedResources, result.ResourceKey)
+			}
+			sc.cacheInvalidationCallback(modifiedResources)
+		}
```
leoluz, line 580: Same comment applies:

Why would we need to invalidate the cache on every sync if the cluster cache is based on resource watches? I feel that the root problem is elsewhere. WDYT?

leoluz (changes requested):
Please check my comments

pkg/engine/engine.go
```diff
@@ -74,6 +74,13 @@ func (e *gitOpsEngine) Sync(ctx context.Context,
 	namespace string,
 	opts ...sync.SyncOpt,
 ) ([]common.ResourceSyncResult, error) {
+	// Ensure cache is synced before getting managed live objects
+	// This forces a refresh if the cache was invalidated
+	err := e.cache.EnsureSynced()
```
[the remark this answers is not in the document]
pjiang-dev: after looking at argo-cd code more it seems like `GetManagedLiveObjs() `actually calls `c.getSyncedCluster(destCluster)` which eventually calls `clusterCache.EnsureSynced()` anyway.

So this part is definitely not needed

PR #747, opened 2025-07-16
fix: update normalization for ignoreDifferences on server-side diff

pjiang-dev:
fixes https://github.com/argoproj/argo-cd/issues/17362#issuecomment-2717284475

When ignoreDifferences normalization removes fields before server-side apply, it causes validation errors because required fields (like image) are missing from the config sent to the Kubernetes API.

To avoid this issue we will
1. Skip the normalizer (ignoreDifferences) with a new option `applyIgnoreDifference` when performing server-side apply --dry-run
2. After predictedLive is performed, we can now normalize (ignoreDifferences) the config and live
3. Result will be a diff without those ignoreDifferences fields

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -851,9 +851,9 @@ func Normalize(un *unstructured.Unstructured, opts ...Option) {
 		normalizeEndpoint(un, o)
 	}
 
-	// Skip full normalization for server-side diff (only apply basic normalization)
-	// For ignoreDifferences to work properly, it needs to be applied during the final
-	// comparison phase for server-side diff, not during input pre-processing.
+	// Skip the full normalization (ignoreDifferences + knownTypes) for server-side diff
+	// In the case an ignoreDifferences field is required, it needs to be present in the config
+	// before server-side diff is calculated and normalized before final comparison.
 	if !o.skipFullNormalize {
 		err := o.normalizer.Normalize(un)
 		if err != nil {


leoluz (changes requested):
PR is missing tests.

pkg/diff/diff.go
```diff
@@ -194,6 +202,107 @@ func serverSideDiff(config, live *unstructured.Unstructured, opts ...Option) (*D
...
+
+	// Calculate what fields were removed from live by normalization (same pattern as livePatch in normalizeTargetResources)
+	var lookupPatchMeta *strategicpatch.PatchMetaFromStruct
+	versionedObject, err := scheme.Scheme.New(originalConfig.GroupVersionKind())
+	if err == nil {
+		meta, err := strategicpatch.NewPatchMetaFromStruct(versionedObject)
+		if err == nil {
+			lookupPatchMeta = &meta
+		}
+	}
```
leoluz, line 230: what is the reason why we are ignoring the errors in this case?

pkg/diff/diff.go
```diff
@@ -194,6 +202,107 @@ func serverSideDiff(config, live *unstructured.Unstructured, opts ...Option) (*D
...
+
+// applyMergePatch applies the patch to the object
+func applyMergePatch(obj *unstructured.Unstructured, patch []byte, versionedObject any) (*unstructured.Unstructured, error) {
+	originalJSON, err := obj.MarshalJSON()
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal original object: %w", err)
+	}
+	var patchedJSON []byte
+	if versionedObject == nil {
+		patchedJSON, err = jsonpatch.MergePatch(originalJSON, patch)
```
leoluz, line 287: Not sure if patching the resources is a good strategy to address this issue. The whole idea of SSD is to avoid having to manipulate resources state by patches. Can we achieve the same using the structure-merge-diff library?

leoluz (changes requested):

pkg/diff/diff.go
pjiang-dev: In this case if there is an error getMergePatch() will fallback to JSON merging as lookupPatchMeta will be nil.
But yes we don't want to ignore it here, we should at least log it.

pkg/diff/diff.go
pjiang-dev: refactoring this to remove this function entirely in favor of simply doing ignoreDifferences (o.Normalizer.normalize) after server-side diff --dry-run is performed, but before final comparison is done.

pkg/diff/diff.go
```diff
@@ -845,9 +845,14 @@ func Normalize(un *unstructured.Unstructured, opts ...Option) {
...
-	if err != nil {
-		o.log.Error(err, fmt.Sprintf("Failed to normalize %s/%s/%s", un.GroupVersionKind(), un.GetNamespace(), un.GetName()))
+	// Skip the normalizer (ignoreDifferences + knownTypes) for server-side diff
+	// In the case an ignoreDifferences field is required, it needs to be be present in the config
+	// before server-side diff is calculated and normalized before final comparison.
+	if o.applyIgnoreDifferences {
+		err := o.normalizer.Normalize(un)
+		if err != nil {
+			o.log.Error(err, fmt.Sprintf("Failed to normalize %s/%s/%s", un.GroupVersionKind(), un.GetNamespace(), un.GetName()))
+		}
```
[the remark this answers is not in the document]
pjiang-dev: What this is doing is skipping this public Normalization function before calculating predicted live. Then after predicted live is calculated, we do the public Normalize: https://github.com/argoproj/gitops-engine/blob/41d976cec906c5e7c24283b644be47eb376d8970/pkg/diff/diff.go#L180 which was already present.

I don't think we can fully skip the Normalization for server-side diff still as it would then show diffs for ignoreDifferences fields.

pkg/diff/diff.go
```diff
@@ -77,11 +77,11 @@ func Diff(config, live *unstructured.Unstructured, opts ...Option) (*DiffResult,
 	o := applyOptions(opts)
 	if config != nil {
 		config = remarshal(config, o)
-		Normalize(config, opts...)
+		Normalize(config, append(opts, WithApplyIgnoreDifferences(false))...)
 	}
 	if live != nil {
 		live = remarshal(live, o)
-		Normalize(live, opts...)
+		Normalize(live, append(opts, WithApplyIgnoreDifferences(false))...)
```
agaudreault, line 84: This might be causing a regression with the 2-way diff and 3-way diff below in this method. When 3-way diff is called, the Normalization with IgnoreDifference=true will be done before calling ThreeWayDiff only for the obj in the last-applied-configuration. However, Normalize is not called anymore for live and config, so `o.normalizer.Normalize(un)` will never be called anymore for 2-way and 3-way diffs. While `ServerSideDiff` does call Normalize again for live and config, it is not the case for the TwoWayDiff/ThreeWayDiff method.
```
	Normalize(orig, opts...)
	dr, err := ThreeWayDiff(orig, config, live)
```

```
	return TwoWayDiff(config, live)
```

My suggestion would be to call normalize based on serverSideDiff option, and preserve the behaviour for other kind of diff. I assume RespectIgnoreDifference is important for 2-way and 3-way diff, and I feel tests are missing for this.

```go
	o := applyOptions(opts)
	preDiffOpts := opts
        if o.serverSideDiff {
		preDiffOpts = append(preDiffOpts, WithApplyIgnoreDifferences(false))
	}
	if config != nil {
		config = remarshal(config, o)
		Normalize(config, preDiffOpts...)
	}
	if live != nil {
		live = remarshal(live, o)
		Normalize(live, preDiffOpts...)
	}
```

PR #749, opened 2025-07-27
chore: added common disable sync options

reggie-k:


pkg/sync/common/types.go
```diff
@@ -28,6 +28,8 @@ const (
 	SyncOptionPruneLast = "PruneLast=true"
 	// Sync option that enables use of replace or create command instead of apply
 	SyncOptionReplace = "Replace=true"
+	// Sync option that disables use of replace or create command instead of apply
+	SyncOptionDisableReplace = "Replace=false"
```
pasha-codefresh, line 32: Could you please explain why you need it? Where you will use it?

reggie-k: Will use it when setting the CLI options to `Replace=false` and `ApplyOutOfSyncOnly=false`

pasha-codefresh (approved):

PR #754, opened 2025-08-12
fix(hooks): always remove finalizers

agaudreault:
Closes https://github.com/argoproj/argo-cd/issues/23226

### Description

The new finalizer on hooks allow us to observe the hook state before Kubernetes tries to delete it. Multiple issue where found causing the finalizers not to be properly removed on hooks managed by the sync. 

- When using SyncFail phase hooks, if any hook fail to apply (sometimes happens with Replace=true or bad manifest), the finalizers will not be removed and the sync will be failed with ongoing running tasks. This is a problem in the gitops-engine logic.
- When any resource fail to apply during a phase, the finalizers wont be removed from the existing hooks of that phase
- When there is a "sync wave hook" error, the finalizers wont be removed from the existing hooks of that phase

### Changes

- ensure hook finalizers are always removed so hooks don’t get stuck by removing finalizers on any existing hook task before creating/starting a new one, and ensure the finalizer is removed before deleting `BeforeHookCreation` hooks.
- Don’t preemptively fail a sync if failure hooks fail to apply. Wait for the failure hooks to complete before Failing. Otherwise, Argo CD will immediately retry while the SyncFail hooks are running.
- Terminate ongoing hooks on errors/terminate
- addresses a race between BeforeHookCreation and Replace=true.
- Clarify docs
- Add test for all finalizers termination case

### Tests

- [x] Test sync all Delete policies & finalizers removed
- [x] Test sync existing hook with finalizers succeeds
- [x] Test sync running + terminate remove all finalizers and running hooks
- [x] Test sync fail + syncFailTasks
- [x] Test sync fail + syncFailTasks fail
- [x] test syncWaveHook fails with syncFailTasks
- [x] test race condition with BeforeHookCreation and Replace=true

### DIFF
--- a/pkg/sync/sync_context.go
+++ b/pkg/sync/sync_context.go
@@ -647,6 +647,18 @@ func (sc *syncContext) Sync() {
 func (sc *syncContext) Terminate() {
 	sc.log.V(1).Info("terminating")
 	tasks, _ := sc.getSyncTasks()
+
+	// Remove completed hook finalizers
+	hooksCompleted := tasks.Filter(func(task *syncTask) bool {
+		return task.isHook() && task.completed()
+	})
+	for _, task := range hooksCompleted {
+		if err := sc.removeHookFinalizer(task); err != nil {
+			sc.setResourceResult(task, task.syncStatus, common.OperationError, fmt.Sprintf("Failed to remove hook finalizer: %v", err))
+		}
+	}
+
+	// Terminate running hooks
 	terminateSuccessful := sc.terminateHooksPreemptively(tasks.Filter(func(task *syncTask) bool { return task.isHook() }))
 	if terminateSuccessful {
 		sc.setOperationPhase(common.OperationFailed, "Operation terminated")
--- a/pkg/sync/sync_context_test.go
+++ b/pkg/sync/sync_context_test.go
@@ -2534,8 +2534,8 @@ func TestTerminate_Hooks_Running(t *testing.T) {
 			metav1.Now(),
 		))
 	syncCtx.resources = groupResources(ReconciliationResult{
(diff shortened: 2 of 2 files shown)


pkg/sync/sync_context.go
```diff
@@ -439,7 +457,7 @@ func (sc *syncContext) Sync() {
 			// No need to perform a dry-run on the namespace creation, because if it fails we stop anyway
 			sc.log.WithValues("task", nsCreateTask).Info("Creating namespace")
 			if sc.runTasks(nsSyncTasks, false) == failed {
-				sc.setOperationFailed(syncTasks{}, nsSyncTasks, "the namespace failed to apply")
+				sc.executeSyncFailPhase(syncTasks{}, nsSyncTasks, "the namespace failed to apply")
```
agaudreault, line 460: `setOperationFailed` was renamed to `executeSyncFailPhase`

pkg/sync/sync_context.go
```diff
@@ -581,33 +594,76 @@ func (sc *syncContext) Sync() {
...
+	tasks, _ := sc.getSyncTasks()
+	terminateSuccessful := sc.terminateHooksPreemptively(tasks.Filter(func(task *syncTask) bool { return task.isHook() }))
+	if terminateSuccessful {
+		sc.setOperationPhase(common.OperationFailed, "Operation terminated")
+	} else {
+		sc.setOperationPhase(common.OperationError, "Operation termination had errors")
+	}
+}
+
+func (sc *syncContext) GetState() (common.OperationPhase, string, []common.ResourceSyncResult) {
```
agaudreault, line 658: Moved GetState(), no changes

PR #759, opened 2025-08-19
fix: call validate on all options to prevent from errors

agaudreault:
The code is simple, but unit test framework is not setup to test the KubectlResourceOperation struct. 

This struct is mocked by the sync context tests.

The unit tests will come in another PR (https://github.com/argoproj/gitops-engine/pull/758), but we can ship this fix first.


pjiang-dev (approved):
LGTM

crenshaw-dev (approved):
Argo CD tests pass in CI.

I also tested `TestJobHookPreSyncForceReplace` from https://github.com/argoproj/argo-cd/pull/23697 locally, and this PR fixes the test failure.

PR #771, opened 2025-09-02
Rebase of #503

dprotaso:
This is a rebase of https://github.com/argoproj/gitops-engine/pull/503

Fixes: https://github.com/argoproj/gitops-engine/issues/501

### DIFF
--- a/pkg/sync/common/types.go
+++ b/pkg/sync/common/types.go
@@ -50,6 +50,8 @@ const (
 	SyncOptionClientSideApplyMigration = "ClientSideApplyMigration=true"
 	// Sync option that disables client-side apply migration
 	SyncOptionDisableClientSideApplyMigration = "ClientSideApplyMigration=false"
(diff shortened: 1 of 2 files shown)


pkg/sync/common/types.go
```diff
@@ -50,6 +50,8 @@ const (
 	SyncOptionClientSideApplyMigration = "ClientSideApplyMigration=true"
 	// Sync option that disables client-side apply migration
 	SyncOptionDisableClientSideApplyMigration = "ClientSideApplyMigration=false"
+	// Sync option that means only controller owner references are respected
+	SyncOptionControllerReferencesOnly = "ControllerReferencesOnly=true"
```
dprotaso, line 54: I'm curious how this options can become the default. Is there a concept in the Argo project as feature gates.

eg. where it's opt-in and then the flag get's switched to opt out?

PR #774, opened 2025-09-03
Track ownerrefs between cluster-scoped and namespaced objects

### LINKED ISSUE #24379: Namespaced objects missing from hierarchy when owned by cluster-scoped objects
# ArgoCD Bug Report: Cross-Namespace Hierarchy Traversal Issue

**Checklist:**
* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

## Describe the bug

ArgoCD fails to display namespaced resources in the application tree when they have ownerReferences to cluster-scoped resources. This occurs due to a cross-namespace hierarchy traversal limitation in the gitops-engine's `IterateHierarchyV2` function.

**Root Cause:** The `buildGraph` function in `pkg/cache/cluster.go` only processes resources within the same namespace, causing cross-namespace parent-child relationships to be missed. Specifically, when a cluster-scoped resource (e.g., Crossplane ProviderRevision) owns namespaced resources (e.g., Deployment, Service), the namespaced children don't appear in ArgoCD's resource tree.

## To Reproduce

1. Deploy a cluster-scoped resource that creates namespaced children via ownerReferences:

```yaml
# Example: Crossplane ProviderRevision (cluster-scoped parent)
apiVersion: pkg.crossplane.io/v1
kind: ProviderRevision
metadata:
  name: provider-aws-cloudformation-3b2c213545b8
---
# Namespaced child with ownerReference to cluster-scoped parent
apiVersion: apps/v1
kind: Deployment
metadata:
  name: provider-aws-cloudformation-3b2c213545b8
  namespace: crossplane-system
  ownerReferences:
  - apiVersion: pkg.crossplane.io/v1
    kind: ProviderRevision
    name: provider-aws-cloudformation-3b2c213545b8
    uid: <provider-revision-uid>
---
# Cluster-scoped child (this WILL appear - works correctly)
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: crossplane:provider:provider-aws-cloudformation-3b2c213545b8:system
  ownerReferences:
  - apiVersion: pkg.crossplane.io/v1
    kind: ProviderRevision
    name: provider-aws-cloudformation-3b2c213545b8
    uid: <provider-revision-uid>
```

2. Create an ArgoCD Application that tracks the ProviderRevision
3. Observe in ArgoCD UI that:
   - ✅ ProviderRevision appears
   - ✅ ClusterRole appears (cluster-scoped → cluster-scoped works)
   - ❌ Deployment is missing (cluster-scoped → namespaced fails)

## Expected behavior

All resources with ownerReferences should appear in the ArgoCD application tree regardless of namespace boundaries. The Deployment should be visible as a child of the ProviderRevision.

## Actual behavior

Only cluster-scoped children appear in the resource tree. Namespaced children of cluster-scoped parents are missing.

## Technical Details

The issue is in `gitops-engine/pkg/cache/cluster.go` at line 1068 in the `IterateHierarchyV2` function:

```go
graph := buildGraph(nsNodes, c.resources) // Should pass c.resources for cross-namespace lookup
```

The `buildGraph` function needs enhancement to handle cross-namespace relationships by:
1. Accepting an `allResources` parameter for global resource lookup
2. Implementing cross-namespace parent resolution logic
3. Processing cross-namespace children in addition to same-namespace resources

## Version

```shell
argocd: v2.12.3+c5b1b3b.dirty
  BuildDate: 2024-08-30T19:14:57Z
  GitCommit: c5b1b3ba3733e68ba954b17d73816ea0273e6cc6
  GitTreeState: dirty
  GoVersion: go1.23.1
  Compiler: gc
  Platform: darwin/arm64
```

## Impact

This bug affects any Kubernetes operator that uses cluster-scoped resources as parents of namespaced resources, including:
- Crossplane providers and configurations
- cert-manager cluster issuers
- Custom operators following similar patterns

## Proposed Fix

The fix requires enhancing the `buildGraph` function in gitops-engine to support cross-namespace hierarchy traversal. A working implementation is available that:
- Maintains backward compatibility
- Adds minimal performance overhead
- Includes comprehensive test coverage
- Passes all existing regression tests

The fix involves updating `pkg/cache/cluster.go` to pass the complete resource map to `buildGraph` and enhancing the function to handle cross-namespace parent-child relationships.

## Logs

No specific error logs are generated - the resources are simply omitted from the tree without warnings or errors.

jcogilvie:
Fixes https://github.com/argoproj/argo-cd/issues/24379

### DIFF
--- a/pkg/cache/cluster.go
+++ b/pkg/cache/cluster.go
@@ -3,6 +3,7 @@ package cache
 import (
 	"context"
 	"fmt"
+	"os"
 	"runtime/debug"
 	"strings"
 	"sync"
@@ -57,6 +58,15 @@ const (
 	defaultListSemaphoreWeight = 50
 	// defaultEventProcessingInterval is the default interval for processing events
 	defaultEventProcessingInterval = 100 * time.Millisecond
+)
+
+// Environment variables
+const (
+	// EnvDisableClusterScopedParentRefs disables cluster-scoped parent owner reference resolution
+	// when set to any non-empty value. This provides an emergency fallback to disable the
+	// cluster-scoped to namespaced hierarchy traversal feature if performance issues are encountered.
+	// Default behavior (empty/unset) enables cluster-scoped parent owner reference resolution.
+	EnvDisableClusterScopedParentRefs = "GITOPS_ENGINE_DISABLE_CLUSTER_SCOPED_PARENT_REFS"
 )
 
 const (
@@ -187,6 +197,8 @@ func NewClusterCache(config *rest.Config, opts ...UpdateSettingsFunc) *clusterCa
 		listRetryLimit:          1,
 		listRetryUseBackoff:     false,
 		listRetryFunc:           ListRetryFuncNever,
+		// Check environment variable once at startup for performance
+		disableClusterScopedParentRefs: os.Getenv("GITOPS_ENGINE_DISABLE_CLUSTER_SCOPED_PARENT_REFS") != "",
 	}
 	for i := range opts {
 		opts[i](cache)
@@ -245,6 +257,10 @@ type clusterCache struct {
 	gvkParser                   *managedfields.GvkParser
 
 	respectRBAC int
+
+	// disableClusterScopedParentRefs is set at initialization to cache the environment variable check
+	// When true, cluster-scoped parent owner reference resolution is disabled for emergency fallback
+	disableClusterScopedParentRefs bool
 }
 
 type clusterCacheSync struct {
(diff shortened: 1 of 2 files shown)


todaywasawesome (commented):
Looks great, as was suggested in contrib meeting we should have an environmental variable. I think this should be disabled by default and only enabled when needed.

jcogilvie:
I heard some other opinions in the contributor meeting.  I agree with adding the escape hatch, but I'm not sure whether we should implement "conditional enable" or "emergency disable".  Environment variable?  App annotation?  Other thoughts?  

@crenshaw-dev had an opinion I know.

jcogilvie:
What's next for this?

PR #776, opened 2025-09-05
fix(sync): ApplyOutOfSyncOnly=true sync option is not honoured for cluster scoped resources (#765)

anandf:
cherry-pick of #765


agaudreault:
Will be part of https://github.com/argoproj/argo-cd/pull/24418

